PR #13, opened 2015-09-23
Add week 8 content

KaranKamath:


.gitignore
```diff
@@ -0,0 +1 @@
+.DS_Store
```
damithc: Remove

damithc:
Minor change requested.

KaranKamath:
Removed and rebased.
Is this because .gitignore should be updated in its own fix, or a standards violation?

damithc:
My comment was not about `.gitignore` file. I was referring to a hidden Mac file (`.DS_Store` or something like that)

damithc:
Actually, my mistake. I confused file name with file content.

PR #35, opened 2015-10-05
 Schedule page '+' buttons can go into an inconsistent state #21 

### LINKED ISSUE #21: Schedule page '+' buttons can go into an inconsistent state
It's possible to go to this state by clicking on the '+' button first and then clicking on the panel heading
image

benjaminheng:
Fixes #21 by checking the expanded state of the clicked accordion and showing/hiding the buttons accordingly.


damithc:
@wkurniawan07 here's another one to review.

wkurniawan07:
Looks good. Only one thing: what's the diff on the very last line about? Can it be excluded? Actually your other PR also has this problem (well not exactly a problem though)

benjaminheng:
My guess is that it's due to different line endings being used. For what
it's worth I used gvim on Windows to edit the file.

damithc:
The fix doesn't solve the problem entirely. Consider this:
1. Click + to see overview
2. Click ++ to expand fully
3. Click - to collapse

Expected: ++
Actual: --

image

benjaminheng:
The buttons should be more consistent now. The button state (`-` or `+`) is determined by the state of the child accordions. For instance, since:

- `+` button shows an overview (expand only immediate children)
- `++` button expands all children

Then if any immediate child is not expanded, the button state will be `+` instead of `-`, and likewise for `++`. Basically, the expand functionality of the button is retained if it is relevant.

scripts/common.js
```diff
@@ -53,6 +54,48 @@ function addCollapseAndExpandButtonsForComponents(accordionHeaderSelector, divId
...
+}
+
+function showExpandAllButton(accordionHeaderSelector) {
+        $(accordionHeaderSelector + ' > .btn-collapseall').hide();
+        $(accordionHeaderSelector + ' > .btn-expandall').show();
+}
+
+function showCollapseAllButton(accordionHeaderSelector) {
+        $(accordionHeaderSelector + ' > .btn-collapseall').show();
+        $(accordionHeaderSelector + ' > .btn-expandall').hide();
```
wkurniawan07: I really don't want to be picky about indentations, but shouldn't this be 4 spaces?

wkurniawan07:
One minor issue then can go ahead and merge.

benjaminheng:
Oh right! Fixed it. Don't worry, that's not being picky at all! :smile:

PR #49, opened 2015-10-13
Animation may not be necessary #45

### LINKED ISSUE #45: Animation may not be necessary
That 0.1 second of waiting time is somewhat frustrating, especially because there are multiple levels to open. Wouldn't faster or instant open be better?

RaghavRamesh:
Fixes #45

### DIFF
--- a/scripts/common.js
+++ b/scripts/common.js
@@ -1,6 +1,7 @@
 function makeAccordion(elementSelector) {
     $(elementSelector).accordion({
         active: false,
+        animate: 0,
         collapsible: true,
         heightStyle: 'content'
     });


damithc:
Good job figuring out an easy way to test the animation @RaghavRamesh 
I've tried it. While it is not too jarring without the animation, for now I'd rather keep the animation because it doesn't take long and is meaningful. I'm going to close this PR for now, but we might revisit this issue again in a future time.

damithc:
Thanks for the contribution @RaghavRamesh 
Feel free to add your name to CONTRIBUTORS.md

Issue #29, opened 2015-09-30
Load weeks' content asynchronously


yj-soh:
The weeks loading is done in a non-asynchronous manner.
This effectively disables the page until everything is loaded on initial load as the JS thread is blocked (preventing the reveal animation for the contents).

PR #62, opened 2015-11-07
Add benjaminheng to CONTRIBUTORS.md

benjaminheng:
Add name to CONTRIBUTORS.md for closing the following two issues: #21 #29 

(Also may I fix the formatting in this file? Namely fixing the typo "schedle" and removing the trailing period in line 3.)


wkurniawan07:
Try not to do anything else outside of what the PR is about. You can open an issue for those and fix it right away.

PR #67, opened 2015-11-16
Convert 'points to ponder' doc to html #51

### LINKED ISSUE #51: Convert 'points to ponder' doc to html
The current page is here https://docs.google.com/document/d/1_4r9JmKvXkcqfATgNe3xPvRHnM32QuJOu4DfegUpd7I/pub?embedded%3Dtrue

This doc is linked from T9 description.

crispyfridge:
Fixes #51 
Created pointsToPonder.html in contents folder and updated relevant link in week10.html


damithc:
Thanks for the PR @crispyfridge 
Some changes to do:
Limit line length to about 110-120 chars.
Separate the css into another file.
Avoid using iframes.

crispyfridge:
@damithc I've made the changes. Please review, thank you.

styles/document.css
```diff
@@ -0,0 +1,32 @@
+body {
```
damithc: Give a more suitable name to the file.

contents/pointsToPonder.html
```diff
@@ -0,0 +1,69 @@
+
```
damithc: Let's name this file pointsToPonderAboutProductDesign.html

contents/pointsToPonder.html
```diff
@@ -0,0 +1,69 @@
...
+        </td>
+      </tr>
+    </tbody>
+  </table>
+  <hr>
+  <p >
+    <span>Here are some things to think about when refining the design of the product.</span></p>
+    <h1><a name="General"></a>
+    <span>General</span></h1>
+    <ul><li ><p class="word-limit" >Make common things easy. What are the most common things a user may want to achieve using the app? e.g. To find out the schedule for today/tomorrow. Make those things VERY easy to do.</p></li></ul>
```
damithc: I did not mean limit display length to 120 chars. I meant in the source code. Just break the lines at appropriate positions so that the code can be viewed without scrolling horizontally.

contents/pointsToPonder.html
```diff
@@ -0,0 +1,69 @@
...
+    <ul><li><p class="word-limit" >Prefer typing over keyboard shortcuts: Keyboard shortcuts are useful and they do not require the mouse. But keep in mind that Jim prefers typing, not necessarily keyboard shortcuts. Keyboard shortcuts involving key combinations often takes more time than regular typing.<br>Automatically generating a command template to fill in may not be good either. For a fast typist, editing an existing text by moving around using arrow/tab keys is slower than typing everything in one burst. <br>The goal is to give Jim a faster &#145;typing&#146; alternative to anything that can be done in other ways. </p></li></ul><p ></p>
+    <ul><li><p class="word-limit" >Avoil multi-step CLIs: The user should be able to accomplish most tasks using a single command. For example, the approach &#145;type add, press enter, then type description, press enter again etc. is unacceptable. </p></li></ul><p ></p>
```
damithc: Get rid of special chars such as `&#146;`.  Use the actual char instead.

contents/pointsToPonder.html
```diff
@@ -0,0 +1,69 @@
...
+      <li class="word-limit" >What about conversions from one task type to another?</li>
+      <li class="word-limit" >Is it better to show deadlines and timed tasks together or separately? Do people process those two types of things separately when they think about things they have to do?</li>
+      <li class="word-limit" >What is the best way to handle overdue tasks? Automatically hide or highlight?</li>
+      <li class="word-limit" >Is there a need to ask for confirmation when an undo feature is available?</li>
+      <li class="word-limit" >How useful is automatic scheduling? Would you follow a schedule automatically generated by an app?  </li>
+    </ul>
+    
+    <div id="footer"><span>Published by <a target="_blank" title="Learn more about Google Drive" href="//docs.google.com/">Google Drive</a></span><span class="dash">&ndash;</span><a href="//docs.google.com/abuse?id=1_4r9JmKvXkcqfATgNe3xPvRHnM32QuJOu4DfegUpd7I">Report Abuse</a><span class="dash">&ndash;</span><span>Updated automatically every 5 minutes</span></div>
```
damithc: There is no need for this line right?

damithc:
Added some more comments.

crispyfridge:
Hi, would it be better if I removed the css file instead? I had thought it was part of the contents and kept it, but now it seems that it was generated by google.
If I were to keep it, would renaming it to pointsForProductDesign.css be better?

damithc:
Yes, can remove if we don't need it. If we keep it, name to match the html file.

crispyfridge:
Hi, I've made the relevant changes and removed the css file.
The line breaks were included, but adding line breaks to image sources broke the image so I didn't insert line breaks for those.

contents/pointsToPonderAboutProductDesign.html
```diff
@@ -0,0 +1,139 @@
...
+          <p ><span ><a href="#Other">Other</a></span></p>
+          <span style="overflow: hidden; display: inline-block; margin: 0.00px 0.00px; border: 0.00px solid #000000; 
+		  transform: rotate(0.00rad) translateZ(0px); -webkit-transform: rotate(0.00rad) translateZ(0px); width: 199.50px; 
+		  height: 119.70px;"><img alt="" 
+		  src="https://lh3.googleusercontent.com/Tlcrf3o-tGf1uy0EKlqPcUSGSxOUbiI9gXt4IwGmYYZzBlFXVerpj_eQIJuNxGzhOkr2OE_1J6FmJgl-w7mtNpJ0VtshQJeAWi2XHW1oS7tDWrKy4cDmRHqEaL6Exl-wcDp5_fk" 
+		  style="width: 199.50px; height: 119.70px; margin-left: 0.00px; margin-top: 0.00px; transform: rotate(0.00rad) 
+		  translateZ(0px); -webkit-transform: 
+		  rotate(0.00rad) translateZ(0px);" title=""></span></p>
+          [<a class="c9" 
+		  href="https://www.google.com/url?q=http://whitenoise.gizmodo.com/i-cant-stop-laughing-at-these-funny-computer-error-mes-1091455400&amp;sa=D&amp;usg=AFQjCNEMLRPyTrFwfwkqtkJQC-U7A5VZFw">
```
damithc: This URL seem to have extra prefix/suffix, probably auto-added by google docs. Get rid of them.

contents/pointsToPonderAboutProductDesign.html
```diff
@@ -0,0 +1,139 @@
...
+  <div id="contents">
+    <table cellpadding="0" cellspacing="0" class="c6">
+      <tbody>
+        <tr ><td colspan="1" rowspan="1">
+          <p ><span ><a href="#General">General</a></span></p>
+          <p ><span ><a href="#Command">Command format</a></span></p>
+          <p ><span ><a href="#UI">UI</a></span></p>
+          <p ><span ><a href="#Search">Search</a></span></p>
+          <p ><span ><a href="#Other">Other</a></span></p>
+          <span style="overflow: hidden; display: inline-block; margin: 0.00px 0.00px; border: 0.00px solid #000000;
```
damithc: Get rid of this fancy formatting. Just a simple image tag is enough.

contents/pointsToPonderAboutProductDesign.html
```diff
@@ -0,0 +1,139 @@
...
+    <h1><a name="UI"></a>UI</h1>
+    <ul><li><p>Optimize the display format:</p></li>
+    <ul><li><p>For example, listing of tasks should be nicely laid out, sorted in a meaningful way,
+	aligned properly, etc so that the user can read the information quickly. For example, the text UIs below have similar 
+	content but one is much more readable than the other.</p>
+      <span style="overflow: hidden; display: inline-block; margin: 0.00px 0.00px; border: 0.00px solid #000000; transform: 
+	  rotate(0.00rad) translateZ(0px); -webkit-transform: rotate(0.00rad) translateZ(0px); width: 562.67px; height: 405.33px;">
+	  <img alt="" 
+	  src="https://lh6.googleusercontent.com/JZk9dkQ4R-EqxkjtBxuXRrLnN59TZXzGfe4oQFWgUH-517l99Zxd3KmsiXsgQb0HHksNXhX60Dz7EoWkjhVLH2jrWkSslNSRu4BRQOmrOOuUEAGcMW-oT8YALYqyDa892eBRmwk" style="width: 562.67px; height: 405.33px; margin-left: 0.00px;
+	  margin-top: 0.00px; transform: rotate(0.00rad) translateZ(0px); -webkit-transform: rotate(0.00rad) translateZ(0px);" title="">
```
damithc: Same here. use a simple image tag.

contents/pointsToPonderAboutProductDesign.html
```diff
@@ -0,0 +1,139 @@
+
+<!DOCTYPE html>
+<html>
+<head>
+
+</head>
+<body>
```
damithc: There is no need to indent after the `html, head, body` tags.

contents/pointsToPonderAboutProductDesign.html
```diff
@@ -0,0 +1,139 @@
...
+	if the result is shown ‘in context’. e.g. showing the newly added task together with the tasks in that day, but 
+	the new one is highlighted</p></li></ul><p ></p><ul><li>Make use of the welcome screen: The welcome screen (what you display 
+	at the very beginning) can be used to display something that is useful to the user, such as the tasks for the day.</p></li></ul><p ></p>
+    <ul><li><p>Scrolling is a design fail: Yes, the use should be able to scroll when there is more information 
+	to show than the display area allows. It is better if the user rarely need to scroll. You should design such that whatever 
+	the user was looking for is already visible in the display area, without having to scroll. </p></li></ul>
+
+    <h1><a name="Search"></a>Search</h1>
+    <ul><li><p>Give sensible search filters: <br>When searching for a task, what can we expect the user to type?
+	Surely, we cannot expect them to type the original tasks name precisely?<br>When designing search options, start with the kind
```
damithc: These lines are still too long (requires horizontal scrolling). Try to limit to 110 chars (including indentation), 120 max.

contents/pointsToPonderAboutProductDesign.html
```diff
@@ -0,0 +1,139 @@
...
+	  that it takes less space but still easy to understand? Is there any point in displaying year, seconds? Is it better to display
+	  the day as well (e.g. Saturday)?</p></li><li><p>But do be careful about going for a Calendar-like UIs 
+	  (i.e., grids). They are a lot of work to implement and hard to get right. It may be easier to link up with existing calendar
+	  interfaces such as Google Calendar.</p></li></ul><p></p>
+    </ul>
+    <ul><li><p>Give enough feedback to the user: When the user executes a command, the feedback given should
+	be informative. e.g. When giving  feedback after adding a task, ‘Success’ is not good enough. Jim should see the
+	details of the task added so that he can verify if it is indeed the task that he meant to add. <br>It may be even more useful
+	if the result is shown ‘in context’. e.g. showing the newly added task together with the tasks in that day, but 
+	the new one is highlighted</p></li></ul><p ></p><ul><li>Make use of the welcome screen: The welcome screen (what you display
```
damithc: It is good to have a line break near a `<p>` tag.

damithc:
Added some more comments.

crispyfridge:
I've made the relevant changes.

contents/pointsToPonderAboutProductDesign.html
```diff
@@ -0,0 +1,167 @@
...
+    <table cellpadding="0" cellspacing="0" class="c6">
+      <tbody>
+        <tr ><td colspan="1" rowspan="1">
+          <p ><span ><a href="#General">General</a></span></p>
+          <p ><span ><a href="#Command">Command format</a></span></p>
+          <p ><span ><a href="#UI">UI</a></span></p>
+          <p ><span ><a href="#Search">Search</a></span></p>
+          <p ><span ><a href="#Other">Other</a></span></p>
+          <img alt="" 
+		  src="https://lh3.googleusercontent.com/Tlcrf3o-tGf1uy0EKlqPcUSGSxOUbiI9gXt4IwGmYYZzBlFXVerpj_eQIJuNxGzhOkr2OE_1J6FmJgl-w7mtNpJ0VtshQJeAWi2XHW1oS7tDWrKy4cDmRHqEaL6Exl-wcDp5_fk">
```
damithc: May be better to make a local copy of this image?

contents/pointsToPonderAboutProductDesign.html
```diff
@@ -0,0 +1,167 @@
...
+          [<a 
+		  href="https://www.google.com/url?q=http://whitenoise.gizmodo.com/i-cant-stop-laughing-at-these-funny-computer-error-mes-1091455400&amp;sa=D&amp;usg=AFQjCNEMLRPyTrFwfwkqtkJQC-U7A5VZFw">
```
damithc: I think this can be simplified (remove prefix/suffix added by GoogleDocs).
`http://whitenoise.gizmodo.com/i-cant-stop-laughing-at-these-funny-computer-error-mes-1091455400`

contents/pointsToPonderAboutProductDesign.html
```diff
@@ -0,0 +1,167 @@
...
+	 for the possibility of task description containing keywords such as ‘from’ ‘to’. e.g., If the
+	 user wants to add the task ‘watch day after tomorrow’; (i.e., watch the movie named ‘day 
+	 after tomorrow’) to the list, there should be a way to enter it without the software 
+	 interpreting ‘tomorrow’ as the day for the task.</p></li></ul>
+	 
+     <ul><li><p>Stick to users’ language: Using the names mentioned in the requirements (e.g. 
+	 floating tasks) in the user guide or in the app itself might not make sense to users. If you 
+	 want to stick with such terms, at least you should explain the terms to users. Do not make 
+	 users learn new terms unless absolutely 
+	 necessary.</p></li></ul>
```
damithc: Don't format each item as a separate list. Instead, put all the list items inside the same `<ul></ul>` tags.

contents/pointsToPonderAboutProductDesign.html
```diff
@@ -0,0 +1,167 @@
...
+	 Jim uses the software from his office computer. A audio playback of the command is unlikely to 
+	 be useful in such an environment. Imagine how Jim’s co-workers would feel if Jim’s computer 
+	 keeps making audio commentary! </p></li></ul>
+	 
+     <ul><li><p>Avoid half-baked features: When you implement a feature, consider all aspects of it. 
+	 For example, if your extra feature is GCal sync, consider things such as how to store floating 
+	 tasks in GCal, how to sync older items in GCal with your software, how to sync repeating events,
+	 and so on. It is not necessary that you implement all those things. However, you should have 
+	 considered all those things and you should have good reasons why you chose what you implemented 
+	 over what you did not implement. “Oh, we didn’t think of that” is not a good answer. </p></li></ul>
```
damithc: Line length looks much better now. :+1:

contents/pointsToPonderAboutProductDesign.html
```diff
@@ -0,0 +1,167 @@
...
+	</p></li></ul>
+
+    <h1><a name="UI"></a>UI</h1>
+    <ul><li><p>Optimize the display format:</p></li>
+	
+    <ul><li><p>For example, listing of tasks should be nicely laid out, sorted in a meaningful way,
+	aligned properly, etc so that the user can read the information quickly. For example, the text UIs 
+	below have similar content but one is much more readable than the other.</p>
+	  <img alt="" 
+	  src="https://lh6.googleusercontent.com/JZk9dkQ4R-EqxkjtBxuXRrLnN59TZXzGfe4oQFWgUH-517l99Zxd3KmsiXsgQb0HHksNXhX60Dz7EoWkjhVLH2jrWkSslNSRu4BRQOmrOOuUEAGcMW-oT8YALYqyDa892eBRmwk">
```
damithc: Create a local copy instead?

contents/pointsToPonderAboutProductDesign.html
```diff
@@ -0,0 +1,167 @@
...
+    <ul><li><p>Give sensible search filters: <br>When searching for a task, what can we expect the user 
+	to type? Surely, we cannot expect them to type the original tasks name precisely?<br>When designing 
+	search options, start with the kind of things users may want to search for. For example, if Jim is 
+	trying to schedule a lunch appointment, he may want to search for tasks (or free times) he has during 
+	lunch hour for the next few days, but he is unlikely to want to search for tasks scheduled during a 
+	particular time in a particular day. Instead, he can simply display that day’s schedule to find that
+	information.</p></li></ul>
+
+    <h1><a name="Other"></a>Other</h1>
+    <ul><li>Saving things only at exit command is a bad idea; Data will be lost if the the program is
```
damithc: Need a line break after `<ul>`. See http://www.w3schools.com/html/html5_syntax.asp for indentation guide

damithc:
More comments added.

wkurniawan07:
@crispyfridge, I can tell that you didn't do a visual check as you have unicode characters without defining meta charset. Please do a visual check, i.e open the HTML file in a web browser and examine it, in terms of layout etc. Compare it to the Google doc version. Which version do you like better? Which formatting(s) (e.g bold for main sentence, highlights, image size, etc.) do you think can be brought to the HTML version? The HTML version should at least not be worse than the Google doc version.

crispyfridge:
@damithc I've made the changes.

@wkurniawan07 Thanks for pointing it out. I've added the meta charset, didn't know that the special characters could be solved that way >.< 
I actually do a visual check before every commit, so that I can see if any content is missing. The css styles were stripped off in the previous commits, so it looks so terrible compared to the google doc :p 
I intend to work on the styles after the HTML contents get the ok from prof damith.

wkurniawan07:
Well, you learn something new everyday :smile: 
I'll be adding a few more comments. Also, later on if you want to demo the page for review, you can push it to `gh-pages` branch and access it from `github.io`. Something like [what I did here](http://wkurniawan07.github.io/website/).

contents/pointsToPonderAboutProductDesign.html
```diff
@@ -0,0 +1,173 @@
...
+<head>
+<meta charset="utf-8"/>
+</head>
+<body>
+
+<div id="header">CS2103 Project: Points to Ponder</div>
+<table cellpadding="0" cellspacing="0" class="c6">
+  <tbody>
+    <tr>
+      <td colspan="1" rowspan="1">
```
wkurniawan07: I'll be addressing the above 5 lines here.
1. Are `id="header"` and `class="c6"` necessary?
2. You may want to consider removing `cellpadding="0" cellspacing="0"`. In HTML5 they are defined via CSS instead.
3. `colspan="1" rowspan="1"` are both default dimensions for a table cell. They don't need to be defined.

crispyfridge:
@wkurniawan07 Thanks for introducing me to gh-pages :) Was using localhost to test previously
I've removed the tag attributes that you mentioned in the previous comment.

damithc:
@crispyfridge the code looks neat now. Can you follow @wkurniawan07's suggestion and post the link to the github.io page so that I can look at how it looks to viewers.

crispyfridge:
This is the link: http://crispyfridge.github.io/website/
The images and text doesn't look good though.

damithc:
Perhaps you can try to improve those things too?
Image size can be adjusted easily. Other style issues may need applying existing css styles or defining new ones. Take your time. The first issue is a learning experience that takes a long time and many iterations.

crispyfridge:
Ok, I'll work on the images next. Thanks for the encouragement :)

crispyfridge:
I've resized the image and added styles. Please review, thank you.

contents/pointsToPonderAboutProductDesign.html
```diff
@@ -0,0 +1,178 @@
...
+</ul>
+
+<h1><a name="UI"></a>UI</h1>
+<ul>
+  <li><p>Optimize the display format:</p></li>
+  <ul>
+    <li><p>For example, listing of tasks should be nicely laid out, sorted in a meaningful way,
+    aligned properly, etc so that the user can read the information quickly. For example, the text UIs 
+    below have similar content but one is much more readable than the other.</p>
+    <img style="width:562px; height:405px"
```
damithc: What if we set the width only? I think the height will adjust automatically.

contents/pointsToPonderAboutProductDesign.html
```diff
@@ -0,0 +1,178 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8"/>
+  <title>CS2103 Project: Points to Ponder</title>
+  <link rel="stylesheet" type="text/css" href="../styles/pointsToPonderAboutProductDesign.css">
+</head>
+<body>
+
+<div>CS2103 Project: Points to Ponder</div>
```
damithc: This is the page title. But it currently appears as normal text. Try to follow the same style as other pages (e.g. handbook page)

styles/pointsToPonderAboutProductDesign.css
```diff
@@ -0,0 +1,9 @@
+td {
+    padding:5pt 5pt 5pt 5pt;
+    vertical-align:top;
+    width:315px;
+}
+
+#win-err-prompt{
+    text-align: right;
+}
```
damithc: This seem like a style we can define in common.css and reuse in other places.

contents/pointsToPonderAboutProductDesign.html
```diff
@@ -0,0 +1,178 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8"/>
+  <title>CS2103 Project: Points to Ponder</title>
+  <link rel="stylesheet" type="text/css" href="../styles/pointsToPonderAboutProductDesign.css">
```
damithc: Also use common.css? We will be using that file to control the overall style of the entire website.

damithc:
Added some more comments. We want to get this page exactly right because other similar pages can follow this one later. @wkurniawan07 any comments?

wkurniawan07:
Since we're already working on the styles now, I'll repeat what I said before: Compare this version with the Google doc version. Which version do you like better? Which formatting(s) (e.g bold for main sentence, highlights, etc.) do you think can be adapted to enhance the readability? `common.css` has many of these styles which you can re-use.

contents/pointsToPonderAboutProductDesign.html
```diff
@@ -0,0 +1,178 @@
...
+  </tbody>
+</table>
+<hr>
+
+<p>Here are some things to think about when refining the design of the product.</p>
+<h1><a name="General"></a>General</h1>
+<ul>
+  <li><p>Make common things easy. What are the most common things a user may want to achieve 
+  using the app? e.g. To find out the schedule for today/tomorrow. Make those things VERY easy 
+  to do.</p></li>
```
wkurniawan07: `<li><p>content</p></li>` looks unnecessary. `<li>content</li>` should work just fine, right?

crispyfridge:
I've noted, will make the commits on Thursday.
Sorry for the late reply.

crispyfridge:
I've added the changes for most of the comments.
@wkurniawan07 I removed the `<p>` from list items, but the contents are stuck together. I'm not sure if putting back the `<p>` would be better, or adding a style to solve this would be better.

@damithc  For the styles in the created css, would it be fine to copy the contents over? Or would it be better to give it a different name, so that it can be reused, e.g the style for `win-err-prompt`?

Also, I'm not sure if its the max-width for `body` from common.css, but the font for text has gotten a lot smaller and hard to read, I'm not sure how to fix this :|

damithc:
Copying the styles would defeat the purpose. Instead, just use styles in common.css 
`win-err-prompt` is not a style, rather it is the object being styled, so that is not a good name. May be you can define a css class that describe the style (e.g. `.align-right`) in common.css and apply it here. The same style can be used in other places where we need to align right.

damithc:
If the text size is similar to that of other pages (e.g. handbook), then it is fine. We can change the text size for the whole website later, by changing the common.css

crispyfridge:
I've made the style changes.

@damithc The text size is smaller than the other pages, should I leave it alone for now?
The current html page can be viewed here:
http://crispyfridge.github.io/website/contents/pointsToPonderAboutProductDesign.html

damithc:
Let's wait for @wkurniawan07's comments. He may be able to advise you on the css stuff.

wkurniawan07:
I realized that the font size of 14px defined in common.css is always overridden. I think for now you can define the font size of 16px on the document head itself, but treat this as an exception, we need to fix it afterwards.

crispyfridge:
Added font size workaround, and additional changes to make it closer to the google doc.
There are some styles needed that is not in common.css, such as the font for the document (the google doc uses Arial) and some colors for headers. Would it be ok to add it to common.css?

wkurniawan07:
You might need to extend the font size workaround to table contents also.
One more thing, the new page doesn't have to be an exact carbon copy of the google doc page. For me, the version you have right now is good enough to serve its purpose. Let's see if @damithc wants some changes.

damithc:
Yes, it need not be a replica of the GoogleDoc. Looks like we are almost there. No more comments from me. @crispyfridge Let me know when you've done any pending changes.
@wkurniawan07 may be create a new issue to get rid of the font size workaround?

crispyfridge:
@damithc I've extended the workaround to apply to table contents, and there are no more pending changes. The page can be seen here: http://crispyfridge.github.io/website/contents/pointsToPonderAboutProductDesign.html

damithc:
Kudos to your persistence @crispyfridge :+1:

crispyfridge:
@damithc Thanks for being patient with me as well :)

PR #72, opened 2015-12-06
Convert 'testing faq' to html #52

### LINKED ISSUE #52: Convert 'testing faq' to html
Current content is at https://docs.google.com/document/pub?id=1LwAFeBLRyVzdAzZxZJQ8pRMeaa-PKY6VMtUUqGm6wlg&embedded=true

crispyfridge:
Fixes #52 
Link to preview: http://crispyfridge.github.io/website/contents/testingFaq.html

### DIFF
--- a/contents/testingFaq.html
+++ b/contents/testingFaq.html
@@ -62,10 +62,11 @@ method, in addition to checking the status message.</div>
 
 <br>
 <h3><a name="test-private"></a>How to test private methods?</h3>
-<div>For now, make them public. In Java, there is a way to access private methods using reflection. 
-Alternatively, we can omit testing private methods because only public methods are exposed to others. 
-If public methods are working correctly, private methods must be working correctly too.<br>
-A more extensive discussion of this issue can be found here.</div>
+<div>For now, make them public. In Java, there is 
+<a href="http://stackoverflow.com/questions/34571/how-to-test-a-class-that-has-private-methods-fields-or-inner-classes">
+a way to access private methods using reflection</a>. Alternatively, we can omit testing private methods because only public 
+methods are exposed to others. If public methods are working correctly, private methods must be working correctly too.<br>
+A more extensive discussion of this issue can be found <a href="http://www.artima.com/suiterunner/private.html">here.</a></div>
 
 <br>
 <h3><a name="test-reading-user"></a>How to test the reading of user commands?</h3>
--- a/contents/testingFaq.html
+++ b/contents/testingFaq.html
@@ -1,0 +1,114 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8"/>
+  <title>FAQ: Unit Testing</title>
+  <link rel="stylesheet" type="text/css" href="../styles/common.css">
+</head>
+<body>
+<div class="align-right dim">CS2103/T</div>
+<h1>FAQ: Unit Testing</h1>
+<table>
+  <tbody >
+    <tr>
+      <td width="20px"></td>
+      <td>
+        <a href="#test-void">How to test void methods?</a><br>
+        <a href="#is-it-enough">Is it enough to check for the correctness of status message only?</a><br>
+        <a href="#test-private">How to test private methods?</a><br>
+        <a href="#test-reading-user">How to test the reading of user commands?</a><br>
+        <a href="#test-text-display">How to test displaying of text to the user?</a><br>
+        <a href="#method-test-another">Can we use one method in our code to test another method?</a><br>
+        <a href="#why-write-test">Why write tests? Manual testing is much easier.</a><br>
+        <a href="#why-TDD">Why TDD?</a><br>
+      </td>
+    </tr>
+  </tbody>
+</table>
+
+<br>
+<h3><a name="test-void"></a>How to test void methods?</h3>
+<div>Use other means to check the effect of the method. E.g., if you are testing a deleteFile(fileName) 
+method, you could do something like this.
+<div class="four-space-indent">
+  <code>deleteFile(fileName);<br>
+  assertFalse(isFileExists(fileName));</code>
+</div>
+</div>
+
+<br>
+<div>In addition, you can refactor to make methods return values. For example, consider the following 
+method.
+<div class="four-space-indent">
+  <code>void display(){<br>
+    // for each line in file<br>
+         //display line<br>
+  }</code>
+</div>
+It can be refactored like this:
+<div class="four-space-indent">
+  <code>void display(){<br>
+    String content = getFileContent(fileName);<br>
+         print(content);<br>
+  }</code>
+</div>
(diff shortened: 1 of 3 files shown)


contents/testingFaq.html
```diff
@@ -0,0 +1,114 @@
...
+         //display line<br>
+  }</code>
+</div>
+It can be refactored like this:
+<div class="four-space-indent">
+  <code>void display(){<br>
+    String content = getFileContent(fileName);<br>
+         print(content);<br>
+  }</code>
+</div>
```
damithc: In the code examples above, indentation is not correct, in the source code as well as when displayed.

damithc:
@crispyfridge After this one, move on to harder ones involving more css and/or JS, leaving easier ones to new comers.

damithc:
I wonder if we can use Gist to embed formatted code snippets. Can you explore that?

crispyfridge:
@damithc Yeah, gists can be embedded in html files. The code snippet would look like the following:
image
However, the gist created will be hosted on my site, and can only be edited by me. 
Alternatively, I could create an anonymous gist, but an anonymous gist cannot be edited and deleted.

damithc:
It would be inconvenient if the code is in a different repo. See if there is any way to serve formatted code from our own repo. 
Also explore other ways of embedding formatted code. If we figure out a way, we can use it in other places too.

crispyfridge:
There is a way to display formatted code from github in html using jsfiddle. However, there is it rather inconvenient, as a specific folder has to be created for the code to be shared, and the files have to be titled the same way, i.e `<title>.html/js/css`. The jsfiddle code display also includes tabs such as css and javascript, which cannot be modified.

The suggested method to showing formatted code from w3schools is wrapping the code in  `<pre>` and `<code>` tags, and using google/code-prettify script with the tags colours the formatted code.

wkurniawan07:
I don't think it needs a complicated solution. Just wrapping it with `<pre>` preserves the whitespace, and everything else is just a matter of styling.

damithc:
For this particular instance we don't need anything more complicated than `<pre>` tags. But I'm interested in finding a general solution for showing code with automatic syntax coloring. 
@crispyfridge can you point to an example of the 'suggested method' you have mentioned? Does it have automatic syntax coloring?

crispyfridge:
@damithc Yes it has syntax coloring. This is an example page using google/code-prettify:
 https://rawgit.com/google/code-prettify/master/examples/quine.html
It uses the script tag`<script src="https://cdn.rawgit.com/google/code-prettify/master/loader/run_prettify.js"></script>` and including the class prettyprint with the language of the code snippet, within the `pre` tag that wraps the code, like `<pre class="prettyprint lang-html">`
More can be found on https://github.com/google/code-prettify

crispyfridge:
I've added the use of `<pre>` and `<code>` for displaying the formatted code

damithc:
The code-pretty thing looks promising. Shall we try it in this PR?

crispyfridge:
I've added the code-prettify in the code blocks.
Preview: http://crispyfridge.github.io/website/contents/testingFaq.html

contents/testingFaq.html
```diff
@@ -0,0 +1,108 @@
...
+method.
+<pre class="prettyprint"><code class="language-java">void display(){
+    // for each line in file
+         //display line
+}
+</code></pre>
+It can be refactored like this:
+<pre class="prettyprint"><code class="language-java">void display(){
+    String content = getFileContent(fileName);
+         print(content);
```
damithc: Indentation problem?

contents/testingFaq.html
```diff
@@ -0,0 +1,108 @@
...
+<br>
+<div>In addition, you can refactor to make methods return values. For example, consider the following 
+method.
+<pre class="prettyprint"><code class="language-java">void display(){
+    // for each line in file
+         //display line
+}
+</code></pre>
+It can be refactored like this:
+<pre class="prettyprint"><code class="language-java">void display(){
```
damithc: I guess adding a line break just before void makes the html file neater but adds an extra line in the display view?

damithc:
Nice. Does it support a lighter theme?

crispyfridge:
@damithc Adding a line break just before void adds an extra line in the display view. Should I add the line break to make the html file neater?
code-prettify only has one theme where the background is not black, but the overall look is not as great.
Preview: http://crispyfridge.github.io/website/contents/testingFaq.html

damithc:
OK, no need to add a line break. We can stick with the dark theme.
Does it allow formatting a word that occurs inside a paragraph? e.g. a mention of a class/method name?
The only other pending change is the indentation problem mentioned earlier.

damithc:
Looks like it can format words inside a paragraph. Can you apply it to words inside paragraphs that represent code elements, the way you've done it for `CityConnnect.java`?

crispyfridge:
I didn't apply syntax highlighting to `CityConnect.java`, but wrapped the `<code>` tags around it. I can do that for the other words that represent code elements. 

For the indentation, I'm not very sure...Is it better if the `<pre>` tag follows the sentenence '...the following method' without the line break? Or should I give an extra indent to the contents within the `<div>`?

contents/testingFaq.html
damithc: Why this line is indented two levels. It should be the same as the previous line right?

damithc:
Sure, just `<code>` wrapping is enough.
As for indentation, I'm referring to extra indentation in line 50. Added another comment there.

crispyfridge:
Line 50 was formatted that way in the google docs, so I thought it was meant to be that way. :sweat_smile: 
Fixed the indentation for that line, and added the `<code>` tags to code elements in paragraph. Also added some urls that were missing from the html.
Preview: http://crispyfridge.github.io/website/contents/testingFaq.html

damithc:
@wkurniawan07 any further comments?

wkurniawan07:
Just some coding style improvements:

1. `void display() {` instead of `void display(){`
2. `// Comment` instead of `//Comment` (not entirely sure if there's an actual convention on this, but most of the Java codes I see have space in that place)
3. Missing space between `>` and `output` in `TextBuddy.exe < input.txt >output.txt`

Everything else looks ok, I like this library :+1:

crispyfridge:
I've made the improvements.

contents/testingFaq.html
```diff
@@ -0,0 +1,109 @@
...
+
+<br>
+<div>In addition, you can refactor to make methods return values. For example, consider the following 
+method.
+<pre class="prettyprint"><code class="language-java">void display() {
+    // for each line in file
+         // display line
+}</code></pre>
+It can be refactored like this:
+<pre class="prettyprint"><code class="language-java">void display(){
```
wkurniawan07: Errrr... actually there's still one more here... :stuck_out_tongue:

PR #73, opened 2015-12-07
Option to expand all collapsed headers #44

### LINKED ISSUE #18: Add an 'expand all weeks' button to the schedule page
We can have `+` and `++` buttons here for expanding/collapsing all weeks in one shot.
image

### LINKED ISSUE #44: Option to expand all collapsed headers
Option to expand all collapsed headers in the schedule. This helps when we need to search the page for a keyword using the Ctrl-F function :)

zeulb:
Fixes #44
Fixed #18 

Added on preference bar

### DIFF
--- a/scripts/common.js
+++ b/scripts/common.js
@@ -7,7 +7,7 @@ function makeAccordion(elementSelector) {
 }
 
 function getContentUsingAjax(section, elementSelector) {
-    pullContent(section, elementSelector, 'Exract from handbook');
+    pullContent(section, elementSelector, 'Extract from handbook');
 }
 
 function pullContent(section, elementSelector, title) {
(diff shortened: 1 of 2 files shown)


scripts/common.js
```diff
@@ -7,7 +7,7 @@ function makeAccordion(elementSelector) {
 }
 
 function getContentUsingAjax(section, elementSelector) {
-    pullContent(section, elementSelector, 'Exract from handbook');
+    pullContent(section, elementSelector, 'Extract from handbook');
```
damithc: Submit this as a separate PR as it is not related to this issue.

damithc:
I would like to see how it looks. Can you push it to `gh-pages` branch and access it from `github.io`. Something like [this](http://wkurniawan07.github.io/website/).

zeulb:
Ok, link to preview is available [here](http://zeulb.github.io/website/)

damithc:
Nice. @wkurniawan07 for your review please

wkurniawan07:
Good to merge :+1:

damithc:
I think something is missing here. Shouldn't there be a change in an html file, to add the buttons and the check boxes?

zeulb:
The buttons are added dynamically by the javascript. But now that I think about it, the code might look neater if the buttons are added statically on the html file. Should I submit another pull request ?

damithc:
The code you have in the preview is different from the one in the PR. One change (the change that uncomment the checkboxes) has been left out of the PR. So the collapse/expand all buttons don't appear at all.

zeulb:
Yes, I didn't uncomment the preference bar on my pull request, I thought it was purposely hidden.

damithc:
It was. But the new buttons appear in the same area as the preference bar right? So without the preference bar, buttons don't appear? If that is the case, we need to enable the preference bar as well.

zeulb:
Yes, it only appear in the preference bar.

damithc:
Can create another PR to enable the preference bar?

zeulb:
Submitted

PR #97, opened 2015-12-16
Add an HTML coding standard #93

### LINKED ISSUE #93: Add an HTML coding standard
It can be a markdown file.
The content can follow https://docs.google.com/document/d/12PJYbQoqjK-0LzaUuguQ4kGE--eikCcHfwzZDGwFOJ0/pub?embedded=true

HooDeLin:
Fixes #93


contents/coding-standards-html.md
```diff
@@ -0,0 +1,167 @@
+# TEAMMATES HTML Style Guide
```
damithc: Remove TEAMMATES

damithc: Add a hyperlinked table of contents here?

contents/coding-standards-html.md
```diff
@@ -0,0 +1,167 @@
...
+  ```html
+  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+  <meta charset="utf-8">
+  ```
+
+- __General Rules__
+  - Indentation
+  
+    For all code languages, we require indentation to be done via soft tabs (using the space character).
+    Hitting Tab in your text editor shall be equivalent to four spaces.
```
damithc: Revise this rule. Indentation should be 2 spaces for html files. Spaces should be used instead of `tab`

contents/coding-standards-html.md
```diff
@@ -0,0 +1,167 @@
...
+         </div>
+         <div id=”frameBodyWrapper” class="container theme-showcase">
+            stuff...
+         </div>
+         <div id=”frameBottom”>
+            stuff...
+         </div>
+     </body>
+    </html>
+  ```
```
damithc: Make sure indentation is 2 spaces per level. There is no need to indent afer `<head>` and `<body>` tags.

contents/coding-standards-html.md
damithc: There is a separate section on indentation. Use this rule to the same section?

contents/coding-standards-html.md
```diff
@@ -0,0 +1,167 @@
...
+  <input value=computer id='someInput' disabled='disabled' tabindex=2> (single/no quotes)
+  <input value="computer" id="someInput" disabled tabindex="2"> (specify boolean variables in the attributes)
+  ```
+
+## Naming convention
+- Class
+  See Css style guide for this section
+
+- Id
+  lower-case, join-by-hyphen
```
damithc: Format this section better. The line break in the source codes doesn't appear in the actual output.

damithc:
This PR may have to wait till the C++ coding standard document is merged. 
Meanwhile, I've added some comments to work on. Have another look at the output and improve formatting where necessary.

damithc:
C++ coding standard has been merged.
1. Look at the Java/C++ standard .md and .html file and try to follow a similar style where possible.
2. Modify the link from README.md

HooDeLin:
The markdown file has been modified to adhere the format from the C++ coding standards.
The link in the README.md has been changed to the md file url after merging

damithc:
Can you stage this version so that I can check out the html page too?
When the side-by-side comparison of Good vs Bad requires horizontal scrolling, use sequential presentation (i.e. bad followed by good) instead.

HooDeLin:
The staging site is [here](http://hoodelin.github.io/website)
The Good vs Bad are changed to use sequential instead. However, the staging version still wraps them into a table so that the Good and Bad can still be formatted with colors

diff

This is the version where it is not wrapped inside the table

HooDeLin:
If we don't want to be wrapped in a table, we can have a minor tweak on the markdown.js so that the good and bad can still have colors if it is not wrapped in a table

damithc:
@wkurniawan07 for review please.

contents/coding-standards-html.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Document Properties
+- __We will use the HTML5 Doctype and HTML5 features__
+  ```html
+  <!DOCTYPE html>
+  ```
+
+  Do __NOT__ include xml namespace in the document.
+  ```html
+  xmlns="http://www.w3.org/1999/xhtml”
```
wkurniawan07: We might as well spell out the entire bad example here, i.e
```html
<html xmlns="http://www.w3.org/1999/xhtml">
```

contents/coding-standards-html.md
```diff
@@ -0,0 +1,231 @@
...
+  - Markup and inline styling
+  
+    Markup defines the structure and outline of a document and offers a structured content.
+    It is not intended to define the look and feel of the content on the page beyond rudimentary concepts such as headers, paragraphs, and lists.
+    The presentation attributes of HTML have all been deprecated and style should be contained in style sheets.
+    No inline styling should be present in any form.
+
+- __General structure for HTML document__
+  ```html
+  <!DOCTYPE html>
```
wkurniawan07: Don't indent after `<!DOCTYPE html>`.

contents/coding-standards-html.md
```diff
@@ -0,0 +1,231 @@
...
+      </td>
+      <td>
+        <pre lang="html">
+  &lt;p&gt;Stuff
+  &lt;p&gt;More Stuff</pre>
+      </td>
+    </tr>
+  </table>
+
+- Void elements do not have a closing tag. Do not use /> for void elements
```
wkurniawan07: Wrap the `/>` as codes.

contents/coding-standards-html.md
```diff
@@ -0,0 +1,231 @@
...
+  Void elements are: <code>area</code>, <code>base</code>, <code>br</code>, <code>col</code>, <code>command</code>, <code>embed</code>, <code>hr</code>, <code>img</code>, <code>input</code>, <code>keygen</code>, <code>link</code>, <code>meta</code>, <code>param</code>, <code>source</code>, <code>track</code>, <code>wbr</code>
```
wkurniawan07: Let's make this part a bit simpler. Write out just a few example of more commonly used void elements, such as `<br>`, ``, `<link>`, `<meta>`, and provide a link for reference to the full list. [Here](http://www.w3.org/TR/html-markup/syntax.html#void-element) is one.

Something like this (with code blocks and links etc):
Example of void elements are br, img, link, and meta. The full list of HTML void elements can be accessed here.

contents/coding-standards-html.md
```diff
@@ -0,0 +1,231 @@
...
+    We prefer readability over file-size savings when it comes to maintaining existing files.
+    Plenty of whitespace is encouraged, along with ASCII art, where appropriate.
+    There is no need for any developer to purposefully compress HTML or CSS, nor obfuscate JavaScript.
+
+  - Markup and inline styling
+  
+    Markup defines the structure and outline of a document and offers a structured content.
+    It is not intended to define the look and feel of the content on the page beyond rudimentary concepts such as headers, paragraphs, and lists.
+    The presentation attributes of HTML have all been deprecated and style should be contained in style sheets.
+    No inline styling should be present in any form.
```
wkurniawan07: As much as I want to agree with this, sometimes it's inevitable. Rephrase it as `Inline styling should not be used, and should be minimized if its usage is really necessary`.

contents/coding-standards-html.md
```diff
@@ -0,0 +1,231 @@
...
+        stuff...
+      </div>
+      <div id="frameBodyWrapper" class="container theme-showcase">
+        stuff...
+      </div>
+      <div id="frameBottom">
+        stuff...
+      </div>
+    </body>
+    </html>
```
wkurniawan07: This is TEAMMATES' HTML structure, which is different from this website's structure. 

1. Remove all mentions of teammates.
2. Following #132, we now have to include `<meta name="viewport" content="width=device-width, initial scale=1">`. Add it in between the charset and the title.
3. The shortcut icon is not used.
4. There is no hard rule on how the body looks like. Just add a comment mentioning that the body content goes there.
5. The script goes after the body content, just before the `</body>` closing tag, as per the recommendation [here](http://stackoverflow.com/questions/196702/where-to-place-javascript-in-a-html-file).

contents/coding-standards-html.md
```diff
@@ -0,0 +1,231 @@
...
+  xmlns="http://www.w3.org/1999/xhtml”
+  ```
+
+- __Character Encoding__
+  All markup should be delivered as UTF-8, as its the most friendly for internationalization.
+  ```html
+  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+  <meta charset="utf-8">
+  ```
+
```
wkurniawan07: Add a section on __Responsiveness__ here. Mention that to support mobile browsing, this tag needs to be added:
```html
<meta name="viewport" content="width=device-width, initial scale=1.0">
```

contents/coding-standards-html.md
```diff
@@ -0,0 +1,231 @@
...
+      <div id="frameBottom">
+        stuff...
+      </div>
+    </body>
+    </html>
+  ```
+
+## External Files
+- Each webpage should include it's own css file if there is any, as well as the general css file.
+- Each webpage should include it's own javascript file(if necessary), as well as the respective common.js, instructor/student.js, and all the 3rd party js files.
```
wkurniawan07: Don't mention `instructor/student.js`.

wkurniawan07:
Added some comments.

HooDeLin:
Updated based on the comments

README.md
```diff
@@ -29,7 +29,7 @@ If you would like to contribute code, here is the procedure:
      TEAMMATES project:
      [JavaScript](https://docs.google.com/document/d/1gZ6WG6HBTJYHAtVkz9kzi_SUuzfXqzO-SvFnLuag2xM/pub?embedded=true),
      [CSS](https://docs.google.com/document/d/1wA9paRA9cS7ByStGbhRRUZLEzEzimrNQjIDPVqy1ScI/pub), 
-     [HTML](https://docs.google.com/document/d/12PJYbQoqjK-0LzaUuguQ4kGE--eikCcHfwzZDGwFOJ0/pub?embedded=true)
+     [HTML](https://github.com/nus-cs2103/website/blob/master/contents/coding-standards-html.html)
```
wkurniawan07: This will display the HTML file (almost) raw. We need to either point it to the version hosted in the website itself, or just use pure markdown for this. @damithc what do you think?

contents/coding-standards-html.md
```diff
@@ -0,0 +1,230 @@
...
+    
+    <!-- Sample stylesheet inclusion -->
+    <link type="text/css" href="styles/common.css" rel="stylesheet">
+  </head>
+  <body>
+    <!-- Body content goes here -->
+  </body>
+  
+  <!-- Sample javascript inclusion -->
+  <script type="text/javascript" src="/js/jquery-minified.js"></script>
```
wkurniawan07: The script inclusion should be inside the `<body>...</body>`, after the content but before the closing tag.

contents/coding-standards-html.md
```diff
@@ -0,0 +1,230 @@
...
+        <pre lang="html">
+  &lt;p&gt;Stuff
+  &lt;p&gt;More Stuff</pre>
+      </td>
+    </tr>
+  </table>
+
+- Void elements do not have a closing tag. Do not use <code>/></code> for void elements
+  Example of void elements are: <code>br</code>, <code>img</code>, <code>link</code>, <code>meta</code>
+  The full list of HTML void elements can be accessed [here](http://www.w3.org/TR/html-markup/syntax.html#void-element)
```
wkurniawan07: Add `.` to the end of each line to break the sentences. Also, shouldn't wrapping the code blocks with &#96; works just as good as `<code>...</code>`?

contents/coding-standards-html.md
```diff
@@ -0,0 +1,230 @@
...
+- Use [attr="value"] for attribute values
+
+  <table>
+    <tr>
+      <th align="center">Good</th>
+    </tr>
+    <tr>
+      <td>
+        <pre lang="html">
+  &lt;input value="computer" id="someInput" disabled="disabled" tabindex="2"&gt;</pre>
```
wkurniawan07: `id="some-input"` to adhere with the naming convention.

contents/coding-standards-html.md
```diff
@@ -0,0 +1,230 @@
...
+  
+  <table>
+    <tr>
+      <th align="center">Bad</th>
+    </tr>
+    <tr>
+      <td>
+        <pre lang="html">
+  &lt;input value=computer id='someInput' disabled='disabled' tabindex=2&gt;(single/no quotes)
+  &lt;input value="computer" id="someInput" disabled tabindex="2"&gt;(specify boolean variables in the attributes)</pre>
```
wkurniawan07: Here too. It's a section on bad examples alright, but bad naming is not part of the issue in the section.

contents/coding-standards-html.md
```diff
@@ -0,0 +1,230 @@
...
+      <td>
+        <pre lang="html">
+  &lt;p&gt;Student Name&lt;/p&gt;</pre>
+      </td>
+      <td>
+        <pre lang="html">
+  &lt;p&gt; Student Name &lt;/p&gt;</pre>
+      </td>
+    </tr>
+  </table>
```
wkurniawan07: Let's add another good example here. Mention that while trailing whitespaces are not allowed, line breaks and indentations are encouraged if it enhances readability. Something like:
```html
<p>
  ...
  ...
  ...
</p>
```
A 3-4 lines worth of text should be good enough as example. If you need some long text for that, [here](https://en.wikipedia.org/wiki/Lorem_ipsum#Example_text) is one :stuck_out_tongue:

contents/coding-standards-html.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Naming convention
+- Class
+  See Css style guide for this section
+
+- Id
+  lower-case, join-by-hyphen
+
+  ```html
+  instructor-email-1
```
wkurniawan07: Display this as actual usage, e.g `<div id="some-div">`, and change the example because the instructor email example makes sense only if you're in TEAMMATES.

wkurniawan07:
More comments added, we're very close!

contents/coding-standards-html.md
wkurniawan07: Forget what I said about adding space. There's a better solution which shortens the line as well: use line breaks, present them like this:
```html
<!-- description 1 -->
<input ...>

<!-- description 2 -->
<input ...>
```

HooDeLin:
The updates are done

wkurniawan07:
It looks good now, only that I'm not sure how to present it (via HTML or via markdown). Need @damithc's agreement on this. The problem with linking to the HTML file is we have to provide the full link because it is referenced from GitHub, not from the website itself.

damithc:
@wkurniawan07 I'm thinking of creating a new repo for stuff common to all our projects. e.g. coding standards. In that case, we can use github.io of that repo to serve the html page?

wkurniawan07:
It's very possible. In fact, [Google does that](https://github.com/google/styleguide).

README.md
wkurniawan07: @HooDeLin meanwhile, you can redirect to [here](https://cdn.rawgit.com/nus-cs2103/website/master/contents/coding-standards-html.html) first. I haven't tried this in production before, so I'm not sure if there are drawbacks, but we need a working version for this at least for now.
@damithc once this change is made this can be merged.

HooDeLin: @wkurniawan07 The link seems to be not working

wkurniawan07: That is expected. Even the current link won't work if you try. For both of them to work, this PR needs to be merged.
If you want to see how it's going to work (don't use this link!), [here you go](https://cdn.rawgit.com/HooDeLin/website/93-add-html-coding-standards/contents/coding-standards-html.html).

HooDeLin:
The links are updated.

wkurniawan07:
Ok, this is ready to go as well.

PR #88, opened 2015-12-16
Convert the C++ coding standard to a markdown document #83

zhongwei-z:
Fixes #83 
Supersedes #86

PR #100, opened 2015-12-16
Provide a way to display markdown outside GitHub #99

### LINKED ISSUE #99: Provide a way to display markdown outside GitHub
As per title

wkurniawan07:
Fixes #99 
The demo uses .md file from #88 (thanks for this!). [Here](http://wkurniawan07.github.io/website/contents/coding-standards-cpp.html) is how it looks like in website, and [here](https://github.com/wkurniawan07/website/blob/gh-pages/contents/coding-standards-cpp.md) in GitHub.


damithc:
Go ahead and merge. Also help @zhongwei-z to apply this technique in his 'C++ coding standard' PR.

damithc:
There is also the question of when to use markdown and where to use html. Markdown files served using strapdown are slower to load and not suitable for embedding. Perhaps we can use strapdown approach for standalone pages with lot of code examples?

wkurniawan07:
Managed to tap into GitHub's markdown API. Now the HTML version looks (almost?) exactly like its GitHub-hosted version, although there is a limited request of 60 per IP address per hour which should be reasonable. Failing which the user will just be referred to the absolute URL in GitHub.

image

As for strapdown, it's not used anymore now but we can always revisit it if the need comes. The technique from this PR is suitable only for pure markdown file.

damithc:
Why not strapdown? 
I'm not sure if the API limit will be breached because too many requests go through the limited number of IP addresses shared by NUS. 
Also, it is nicer if we can use GFMD, not just pure markdown.

wkurniawan07:
That's exactly why I moved away from strapdown: with GitHub API we can get GFMD, not just any markdown flavor, although the API limit is of course going to be annoying.

damithc:
If the fallback is going to be the repo version, then we should use the repo version all the time? The only inconvenience is we have to use the absolute URL instead of the relative URL, which is not a big deal.

wkurniawan07:
Ok, now the fallback is automatic redirect to the GitHub version (no more error displayed screen). I'll see if I can replicate the GFM without having to call GitHub's API. Until then this method should serve its purpose well.
The main advantage of hosting own HTML file is that we're free to tweak the page's CSS as needed.

damithc:
It is nice if we can tweak styles. Let's say we want to show 'Bad' in red and 'Good' in green (i'm referring to the labels we use for code samples). How would we do that?

wkurniawan07:
Inline styling or adding classes won't work even in GitHub itself because the API will strip them all, so there's pretty much no way to tweak the style in the GitHub version. For HTML version:

1. Inline styling or classes are stripped, but ids are retained (although the API will prepend `user-content-` to it). We can use the id for css selector. (actually this is also the case for GitHub version, but appending id is no use if we can't do anything with it)
2. We can do simple string replaces via js, e.g replace the `Good` with `<span style="color: green;">Good</span>`. We can even involve regexes for global replacement or pattern searching.
3. Alternatively, we can traverse the document with jquery and do what we need.

Here's [a demo](http://wkurniawan07.github.io/website/contents/coding-standards-cpp.html). The table is made borderless with method 1, has its font resized with method 3, and all the goods and bads are colored with method 2. [In GitHub](https://github.com/wkurniawan07/website/blob/gh-pages/contents/coding-standards-cpp.md) this is not possible.

damithc:
Ok, let's try it out. May be generalize it a bit (we have several coding standards documents, and we don't want to duplicate too much code in each of those)?

wkurniawan07:
Minor change made. This is already the most general I can come up with, everything needs to be there. The users of this template only need to change the `markdownUrl` and the pre- and post-process parts of `processAndDisplayResult`.

damithc:
All coding standard docs are going to need the same style tweaks. Can we abstract those in a reusable way?
Perhaps merge the C++ standard first and apply this technique to it in a way that is reusable by other coding standard docs?

wkurniawan07:
That's easy, we can tweak the style on the javascript file itself right after the success function before the `processAndDisplayResult`. Then it'll be reused by all.

wkurniawan07:
We're ready to go. Here's [the C++ version](https://wkurniawan07.github.io/website/contents/coding-standards-cpp.html) and [the Java version](https://wkurniawan07.github.io/website/contents/coding-standards-java.html). Limited to 60 visits per IP address per hour, exceeding which the page will simply be redirected to the GitHub version.

damithc:
Nice. :+1: Go ahead and merge.

PR #104, opened 2015-12-20
Make the whole page scrolls #41

### LINKED ISSUE #41: Make the whole page scrolls
Currently only the div#content scrolls when content overflows (i.e., when sections are expanded).
However, since there are no other content, it's more natural for the whole page to scroll.

HooDeLin:
Fixes #41 
Staging site is [here](http://hoodelin.github.io/website/)

Proposed solution:
The whole page is unable to scroll is because we have fixed the height of the content.
My solution is to check if any accordion is expanded after user clicks on an accordion. The content height is set to 'auto' when all of the accordion is 'collapsed', or fix the content height if there exists an accordion that is expanded.

### DIFF
--- a/scripts/common.js
+++ b/scripts/common.js
@@ -6,6 +6,19 @@ function makeAccordion(elementSelector) {
     });
 }
 
+function makeWeeklyScheduleAccordion(elementSelector) {
+    makeAccordion(elementSelector);
+    $(elementSelector).accordion({
+        activate: function(event, ui) {
+            if(checkIfContentIsExpanded()) {
+                $('#content').css('height', calculateContentContainerSize());
+            } else {
+                $('#content').css('height', 'auto');
+            }
+        }
+    });
+}
+
 function getContentUsingAjax(section, elementSelector) {
     pullContent(section, elementSelector, 'Exract from handbook');
 }
@@ -173,6 +186,19 @@ function checkIfAllComponentsChecked() {
     return isAllChecked;
 }
 
+function checkIfContentIsExpanded() {
+    var isAccordionExpanded = false;
+    $('.weeklyschedule').each(function(i) {         
+        function isCurrentAccordionActive(currentAccordion) {
+            return !(typeof currentAccordion.accordion( "option", "active" ) == "boolean");
+        }
+        if (isCurrentAccordionActive($(this))) {
+            isAccordionExpanded = true;
+        }
+    });
+    return isAccordionExpanded;
+}
+
 function loadContent(week) {
     $.ajax({
         type: 'GET',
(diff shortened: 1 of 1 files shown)


scripts/common.js
```diff
@@ -215,14 +249,14 @@ $(document).ready(function() {
...
-    $('#content').css('height', calculateContainerSize());
+    $('#content').css('height', 'auto');
 
     $(window).resize(function() {
-        $('#content').css('height', calculateContainerSize());
+        if(checkIfContentIsExpanded()) {
+            $('#content').css('height', calculateContentContainerSize());
+        } else {
+            $('#content').css('height', 'auto');
+        }
```
wkurniawan07: Extract this method out, it's used twice.

scripts/common.js
```diff
@@ -173,6 +186,19 @@ function checkIfAllComponentsChecked() {
...
+    var isAccordionExpanded = false;
+    $('.weeklyschedule').each(function(i) {         
+        function isCurrentAccordionActive(currentAccordion) {
+            return !(typeof currentAccordion.accordion( "option", "active" ) == "boolean");
+        }
+        if (isCurrentAccordionActive($(this))) {
+            isAccordionExpanded = true;
+        }
+    });
+    return isAccordionExpanded;
```
wkurniawan07: 1. Don't define the function inside the `.each` loop; define it outside the loop.
2. Use `typeof ... != 'boolean'` instead.
3. Single quotes for js strings.
4. Rather than keeping a variable `isAccordionExpanded`, I think you can just `return true;` if the condition is true, and close the function with `return false;`. This way the function will also terminate directly if needed instead of unnecessarily looping the rest.

scripts/common.js
```diff
@@ -199,9 +225,17 @@ function loadContent(week) {
     });
 }
 
+function calculateContentContainerSize() {
+    var bannerHeight = 25;
+    var headerHeight = 40;
+    var topMargin = 5;
+    var topPadding = 5;
```
wkurniawan07: I think these values are duplicates. See if you can make them defined just once each, e.g with global variables? I don't encourage global variables, but if it's the only way to do this, then so be it.

damithc:
I think the original issue is about this issue. i.e. the scrollbar appears for the content section instead of for the whole page.
image
Does your PR fix it? Based on the staging site, the behaviour seems same as before.

HooDeLin:
In my staging site, the whole page will scroll if there is no expanded content. The scrollbar will resume back to the content if the there is any content expanded.
Original site:
original-site
Staging-site:
modified-site

This can be tested if we have smaller window size.

HooDeLin:
I think that the headings should be left unscrollable since the content can get very long and the user may need to scroll all the way to the top to switch to other documents

damithc:
Yes, the top part should not scroll. 
If the 'whole page scrolling' is applied only when the weeks are collapsed, I don't see much use of it. In most screen sizes it doesn't even need scrolling when everything is collapsed. I think the original poster  was requesting for 'whole page scrolling' even when some weeks are expanded. But I understand if it is hard to do.

HooDeLin:
If that is the case, I believe that it can be easily achieved by not fixing the content size when the document loads. This will make sure that the whole page scrolls

damithc:
Can you submit a separate PR using that approach? We can compare the two and select one.

HooDeLin:
The alternative fix is up on a separate PR already.

PR #105, opened 2015-12-20
Make page fragments previewable #89

### LINKED ISSUE #89: Make page fragments previewable
Currently, page fragments (e.g. `contents\week0.html`) cannot be previewed in a browser. That is, cannot be tested by opening it in a browser. This is because on its own, it does not have a `<head>` section containing links to style files etc. This is an inconvenience when modifying those files. Everytime we do a modification, we have to open the entire schedule page, and navigate to the relevant section to see how it looks.
It is better if the page fragment can be viewed in a browser, on its own.

ravic1995:
Fixes #89 
I am not really sure if my way is modular. 
I wish I could get some help. 
I attach a ```<script></script>``` to all the standalone week pages, the script then injects the ```common.css``` file into the webpage when it loads.

### DIFF
--- a/contents/week0.html
+++ b/contents/week0.html
@@ -74,9 +74,10 @@
 </div>
 
 <script>
-  loadcss = document.createElement('link');
-  loadcss.type = 'text/css';
-  loadcss.rel = 'stylesheet';
-  loadcss.href = '../styles/comon.css';
-  document.getElementById('head')[0].append(loadcss);
+  var head = document.getElementsByTagName("head")[0];         
+  var loadCss = document.createElement('link');
+  loadCss.type = 'text/css';
+  loadCss.rel = 'stylesheet';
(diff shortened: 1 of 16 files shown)


damithc:
@wkurniawan07 for your review please

wkurniawan07:
This issue is not as easy as it looks. It's going to take some steps to completely resolve it, but hey let's do it.
First things first, your method here will load 15 instances of `common.css` imports to the main schedule page. It's doable but I don't think it's really best done this way.
Instead, use [this method](http://api.jquery.com/load/) in place of `$.ajax`. Give an id to the main div wrapper (let's say `div id="main-content"` or something similar), then append `#main-content` to the name of the document loaded. This way we can develop the page fragments as their own HTML pages (with styles and stuffs) instead of simply div fragments.

PR #106, opened 2015-12-20
Convert the Java coding standard into a markdown document #82

### LINKED ISSUE #82: convert the Java coding standard into a markdown document
The current one is given in https://docs.google.com/document/pub?id=1iAESIXM0zSxEa5OY7dFURam_SgLiSMhPQtU0drQagrs&&embedded=true

barnytan:
Fixes #82 
Based on the C++ coding standards md from #88


damithc:
Let's wait until #88 and #100 are merged before merging this.

contents/coding-standards-java.md
```diff
@@ -0,0 +1,729 @@
+# Java Coding Standard (Beginner/Intermediate Level)
+<table>
+  <tr>
+    <td width="70%" >
+      <ul>
+        <li><a href="#naming-conventions">Naming Conventions</a>
```
wkurniawan07: While it is fine to omit `</li>` in some cases, it's a good practice to write it out.

contents/coding-standards-java.md
```diff
@@ -0,0 +1,729 @@
...
+          <li><a href="#conditionals">Conditionals</a>
+        </ul>
+        <li><a href="#layout-and-comments">Layout and Comments</a>
+        <ul>
+          <li><a href="#layout">Layout</a>
+          <li><a href="#white-space">White Space</a>
+          <li><a href="#comments">Comments</a>
+        </ul>
+		<li><a href="#references">References</a>
+        </ul>
```
wkurniawan07: The indentation for the above two lines are slightly off.

contents/coding-standards-java.md
```diff
@@ -0,0 +1,729 @@
...
+    </td>
+    <td width="20%">
+      <img src='https://lh3.googleusercontent.com/EcEMupVi6p4Cw7_D9nqH_kimUENEvgi412Ci56fNXFUplx_u1miw_L0QMPSN7KscNHmtyK74LCMT3NIVCUSW6qh_fiHcVyShNIZfXcdaTfrD7nWWXyzU3qzA5i1JD2usMy9pvpA' style='width: 140px'>
+      <br>
+      <a href="https://www.google.com/url?q=http://en.wikipedia.org/wiki/Ada_Lovelace&sa=D&usg=AFQjCNHDAlAncZBC-bSxfit5yVEhNlT2AA">Ada Lovlace, the first computer programmer</a>
```
wkurniawan07: You can host the image in this repo itself, and also you can remove the google URL redirection (i.e just link to the wiki page directly). Another thing, use double-quotes `"` instead of single-quotes for HTML tag attributes.

contents/coding-standards-java.md
```diff
@@ -0,0 +1,729 @@
...
+  ...
+}
+```
+
+The notation is taken from mathematics where it is an established convention for indicating iterators. Variables named *j*, *k* etc. should be used for nested loops only.
+
+**13. Associated constants (final variables) should be prefixed by a common type name.**
+
+```java
+final int COLOR_RED		= 1;
```
wkurniawan07: There's an extra tab over here.

contents/coding-standards-java.md
```diff
@@ -0,0 +1,729 @@
...
+  }
+}
+```
+Enforced by the Java tools.
+
+**3. Use line wrapping to improve readability.**
+
+When wrapping lines, the main objective is to improve readability. Feel free to break rules if it improves readability. Do not always accept the auto-formatting suggested by the IDE.
+
+Length: A line should be split if it exceeds the ~~80~~ 110 characters. But it is OK to exceed the limit slightly or wrap the lines much shorter than limit.
```
wkurniawan07: We can omit the striked 80, I guess.

contents/coding-standards-java.md
```diff
@@ -0,0 +1,729 @@
...
+       param4);
+void method(param1,param2)throws Exception1,
+                                 Exception2,
+                                 Exception3 {
+                                 
+//OK TO DO THIS BECAUSE THE COLUMN IS WIDE
+longMethod(someLongMenthod1(param1, param2, param3).anotherMethod(),
+           someLongMenthod2(param1, param2).anotherMethod(),
+           someLongMenthod3(param1, param2, param3));
+Here are three acceptable ways to format ternary expressions:
```
wkurniawan07: This is not part of the code :p

contents/coding-standards-java.md
```diff
@@ -0,0 +1,729 @@
...
+
+There are a few alternatives to the is prefix that fits better in some situations. These are *has*, *can* and *should* prefixes:
+
+```java
+boolean hasLicense();
+boolean canEvaluate();
+boolean shouldAbort = false;
+```
+
+Note: avoid boolean variables that represent the negation of a thing. e.g., use `isInitialized` instead of `isNotInitialized`
```
wkurniawan07: Use a blockquote for this, like:
>Note: this is a note

contents/coding-standards-java.md
```diff
@@ -0,0 +1,729 @@
...
+
+Importing classes explicitly gives an excellent documentation value for the class at hand and makes the class easier to comprehend and maintain. Appropriate tools should be used in order to always keep the import list minimal and up to date. For example, Eclipse IDE can do this easily.
+
+### **Classes and Interfaces**
+
+**4. Class and Interface declarations should be organized in the following manner:**
+
+>1. Class/Interface documentation (Comments)
+>2. **class** or **interface** statement
+>3. Class (static) variables in the order **public**, **protected**, **package** (no access modifier), private
```
wkurniawan07: You forgot to bold the `private` here :p

contents/coding-standards-java.md
```diff
@@ -0,0 +1,729 @@
...
+### **Loops**
+
+**9. The loop body should be wrapped by curly brackets irrespective of how many lines there are in the body**
+
+```java
+sum = 0;          
+for (i = 0; i < 100; i++) {                
+  sum += value[i];
+}
+// NOT: for (i = 0, sum = 0; i < 100; i++)
```
wkurniawan07: This way you're commenting part of the code as well. Line break after the `NOT:` statement will do the trick.

contents/coding-standards-java.md
```diff
@@ -0,0 +1,729 @@
...
+
+When there is only one statement in the loop body it can be written without wrapping it between `{ }`, however that is error prone and *very* strongly discouraged from using.
+
+### **Conditionals**
+
+**10. The conditional should be put on a separate line.**
+
+```java
+if (isDone)       // NOT: if (isDone) doCleanup();
+  doCleanup();
```
wkurniawan07: As per what's written just before this, let's wrap it in `{ ... }` as well.

wkurniawan07:
Some general comments:

1. It's good to see 2 spaces indentations followed (well, mostly), but in Java code blocks they'll also be translated as such and we define the indentation to be 4 spaces. You should fix the ones involving code blocks to 4 spaces.
2. Right now there's some variation to how good code examples and bad code examples are presented. It will be good to standardize it. You can try what's done in #88 (i.e with `Good` and `Bad` keywords and using table where appropriate).

Once #88 and #100 are merged I might have more things to add, but for now you can improve on the areas I've suggested.

contents/coding-standards-java.md
barnytan: The extra tab is for the ='s to be aligned. In the original google doc it is shown like this:
capture2
Is it important for the ='s to be aligned? If not I will put a single space instead 
```java
final int COLOR_RED = 1;
final int COLOR_GREEN = 2;
final int COLOR_BLUE = 3;
```

barnytan:
Thanks for the comments, will work on them.

contents/coding-standards-java.md
wkurniawan07: I see what you mean. Then make it like this:
```java
final int COLOR_RED   = 1;
final int COLOR_GREEN = 2;
final int COLOR_BLUE  = 3;
```
i.e make it aligned but don't use tabs.

barnytan: Got it. Thanks!

contents/coding-standards-java.md
```diff
@@ -0,0 +1,945 @@
...
+```
+
+- Prefer higher-level breaks to lower-level breaks. In the example below, the first is preferred, since the break occurs outside the parenthesized expression, which is at a higher level.
+
+**GOOD:**
+```java
+// PREFER THIS
+longName1 = longName2 * (longName3 + longName4 - longName5)
+            + 4 * longname6;
+            +
```
wkurniawan07: Extra `+` here

contents/coding-standards-java.md
```diff
@@ -0,0 +1,945 @@
...
+
+```java
+totalSum = a + b + c 
+          + d + e;
+setText("Long line split"
+         + "into two parts.");
+method(param1,
+           object.method()
+             .method2(),
+           param3);
```
wkurniawan07: I think this part is intended to be like this:
```
method(param1,
       object.method()
             .method2(),
       param3);
```

contents/coding-standards-java.md
```diff
@@ -0,0 +1,945 @@
...
+        statements;
+        break;
+    
+  case XYZ :
+        statements;
+        break;
+    
+  default :
+        statements;
+        break;
```
wkurniawan07: The cases DEF, XYZ, and default statements are slightly off.

contents/coding-standards-java.md
```diff
@@ -0,0 +1,945 @@
...
+>- Empty line between description and parameter section
+>- Alignment of parameter descriptions
+>- Punctuation behind each parameter description
+>-No blank line between the documentation block and the method/class
+
+Javadoc of class members can be specified on a single line as follows:
+
+```java
+    /** Number of connections to this database */
+    private int nConnections;
```
wkurniawan07: We don't need the indentations here.

wkurniawan07:
Added a few more comments.

contents/coding-standards-java.md
barnytan: It the original google docs it is show as such:
capture3

But based on the point before this, "Align the new line with the beginning of the parent element.", yours might be what is intended.

I'll change it to what you mentioned.

contents/coding-standards-java.md
```diff
@@ -0,0 +1,950 @@
...
+
+```java
+totalSum = a + b + c 
+          + d + e;
+setText("Long line split"
+         + "into two parts.");
+method(param1,
+           object.method()
+                 .method2(),
+           param3);
```
wkurniawan07: Indent the above three lines 4 spaces back, i.e make it like this:
```
method(param1,
       object.method()
             .method2(),
       param3);
```

contents/coding-standards-java.md
```diff
@@ -0,0 +1,950 @@
...
+} 
+      </pre>
+    </td>
+    <td>
+      <pre lang="java">
+while (true) {
+    // Do something
+       something();
+}
+      </pre>
```
wkurniawan07: Let's add one more bad example here:
 ```java
while (true) {
// Do something
    something();
}
```

It's up to you whether you want to put it on the same table cell or a new cell to the right of it.

**Comment by wkurniawan07** [comment]
Two small improvements and we're good to go :+1:

**Comment by barnytan** [comment]
Made an additional change, converted all the Good/Bad examples to use the table format for readability.

File: contents/coding-standards-java.md (hunk 16 above)
Comment (wkurniawan07) [inline contents/coding-standards-java.md]: @barnytan the new example is no different than the existing one. What I'm suggesting is the `// Do something` comment is aligned similar to the `while (true)` statement.

File: contents/coding-standards-java.md [hunk 17]
```diff
@@ -0,0 +1,1007 @@
...
+- Break before an operator. This also applies to the following "*operator-like*" symbols: the dot separator `.`, the ampersand in type bounds `<T extends Foo & Bar>`, and the pipe in catch blocks `catch (FooException | BarException e)`
+
+```java
+totalSum = a + b + c 
+          + d + e;
+setText("Long line split"
+         + "into two parts.");
+method(param1,
+       object.method()
+           .method2(),
```
wkurniawan07: We're very close... just 2 spaces to the right for this particular line.

wkurniawan07:
And one last thing... between the last character in `<pre>` and the closing `</pre>` tag there shouldn't be any whitespace/line break because they will be rendered. e.g do this:
```
      <pre lang="java">
exportHtmlSource();
openDvdPlayer();</pre>
```
instead of this, although it's more readable this way:
```
      <pre lang="java">
exportHtmlSource();
openDvdPlayer();
      </pre>
```

barnytan:
Sorry for the mistakes, corrected them

contents/coding-standards-java.md
```diff
@@ -0,0 +1,967 @@
...
+    // Do something
+    something();
+}</pre>
+    </td>
+    <td>
+      <pre lang="java">
+while (true) {
+    // Do something
+       something();
+}</pre>
```
wkurniawan07: One more here. Make it like
```
while (true) {
// Do something
    something();
}
```
i.e the `// Do something` comment should be aligned the same as the `while (true)` statement in this bad example. Otherwise the two bad examples are the same.

barnytan: Ah I see, I misunderstood it, will correct it.

barnytan: In this case, does it matter if the `something()` is aligned to the comment?

wkurniawan07: Use the standard 4 spaces indentation for that. Bad indentation for codes shouldn't be the issue here.

barnytan: Alright corrected it

wkurniawan07:
It's good now. Well done :+1:

barnytan:
Thanks. On a side note, Merry Christmas.

wkurniawan07:
@barnytan same to you!

PR #108, opened 2015-12-21
Handbook TOC links are not working in iPhone browser #19

### LINKED ISSUE #19: Handbook TOC links are not working in iPhone browser
These links don't seem to be working in iPhone browser
image

crispyfridge:
Fixes #19 
It seems that the `handbook.js` interfered with how the links worked, since the links work fine without the js file, but I'm not sure what is causing that. 
I added empty divs with ids that don't overlap with those that the js is using, and it worked fine.
preview: http://crispyfridge.github.io/website/

### DIFF
--- a/contents/handbook.html
+++ b/contents/handbook.html
@@ -22,225 +22,225 @@
 
 
 <ul>
-    <li><a href="#handbook-preliminaries">Preliminaries</a></li>
-    <li><a href="#handbook-lectures">Lectures</a></li>
-    <li><a href="#handbook-tutorials">Tutorials</a></li>
-    <li><a href="#handbook-textBooks">Text Books</a></li>
-    <li><a href="#handbook-programmingLanguages">Programming Language</a></li>
-    <li><a href="#handbook-project">Project</a></li>
+    <li><a href="#preliminaries">Preliminaries</a></li>
+    <li><a href="#lectures">Lectures</a></li>
+    <li><a href="#tutorials">Tutorials</a></li>
+    <li><a href="#textBooks">Text Books</a></li>
+    <li><a href="#programmingLanguages">Programming Language</a></li>
+    <li><a href="#project">Project</a></li>
     <ul>
-        <li><a href="#handbook-theProduct">The product</a></li>
-        <li><a href="#handbook-projectScope">Project Scope</a></li>
-        <li><a href="#handbook-projectConstraints">Project Constraints</a></li>
-        <li><a href="#handbook-deliveables">Deliverables</a></li>
+        <li><a href="#theProduct">The product</a></li>
+        <li><a href="#projectScope">Project Scope</a></li>
+        <li><a href="#projectConstraints">Project Constraints</a></li>
+        <li><a href="#deliveables">Deliverables</a></li>
         <ul>
-            <li><a href="#handbook-ce1">CE1: TextBuddy</a></li>
-            <li><a href="#handbook-ce2">CE2: TextBuddy++</a></li>
-            <li><a href="#handbook-ce3">CE3: TextBuddyPro</a></li>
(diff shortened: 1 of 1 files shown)


damithc:
May be it's worth figuring out what caused the problem, to avoid it recurring or to avoid breaking something while we are trying to fix this one.

crispyfridge:
From what I read [here](http://www.webdeveloper.com/forum/showthread.php?259280-anchor-tag-not-working-in-safari-(ios)-for-Iphone-iTouch-iPad), it seems that this is a problem unique to iOS. Also, I've tested the site on an android device prior the fix and it works fine.

damithc:
I meant how did this change fix the problem?

crispyfridge:
It was a workaround that I found online, but there wasn't any particular explanation to why it worked.

damithc:
Let's wait to see if @wkurniawan07 has any insights about this.

wkurniawan07:
I'm afraid not, I'm not an iPhone user. I do have an iPad though, and even the current version (before this fix) actually works. @crispyfridge where exactly do you find the workaround?

crispyfridge:
@wkurniawan07 [I got it here](http://answers.squarespace.com/questions/63483/why-wont-my-anchor-links-work?page=1#67402) 
At first I didn't really focus on this solution because it seems like that's what the original code is already doing. I eventually added a new div like the solution suggested when other workarounds failed, and it worked.

damithc:
I used my iPhone to access your staged handbook page from direct URL as well as via home page. Links worked in both cases.

crispyfridge:
@damithc Sorry I shouldn't have used the staged website as an example. The staged website contains the fix, so it works. I wanted to say that before the fix, using mobile, the links in the handbook page from the direct URL works, but the links in the handbook page when accessed from the home page doesn't work.

damithc:
Got it :-)

wkurniawan07:
@crispyfridge maybe you can push it to `gh-pages`, and I'll try it on my PC, android phone, as well as iPad. If it works on my devices and works on iPhones (which I can't confirm) as well, this should be good to go.

crispyfridge:
@wkurniawan07 I've pushed it to gh-pages. You can try it here: http://crispyfridge.github.io/website/

damithc:
and position near to top?

-- 
Sent from a mobile device.

damithc:
Can do as a separate issue.

-- 
Sent from a mobile device.

crispyfridge:
@damithc I've created a separate issue for it

wkurniawan07:
Strange, it doesn't work on my iPad. Even the current version (before this fix) fails now, I could've sworn it worked 2 days ago.
I have a feeling it has something to do with how mobile Safari interprets frames, because the links in the standalone handbook page work just fine.

crispyfridge:
@wkurniawan07 Would it be better if I replaced the frame and frameset in `index.html` with `<iframe>` instead? Or should I use divs?

wkurniawan07:
Divs won't work unless all of the pages used are within our control, which are not (e.g we have the IVLE page and some Google docs page). I haven't tried `iframe` myself, feel free to try it and share the result.

damithc:
Any progress? As a last resort, we can implement the the thing that 'seems to work' although we don't understand why it works. It's not ideal but better than nothing :-)

crispyfridge:
I managed to replace the frames with `iframe`, but it doesn't work on my iphone with the original div ids :(

wkurniawan07:
@crispyfridge I haven't checked on this for a while. Any updates?

crispyfridge:
@wkurniawan07 sorry for the late reply. I haven't been working on this for a while now, so no updates

acjh:
I appreciate any help to test the fix in PR #216 please :)

acjh:
Fixed in #216

PR #109, opened 2015-12-21
Make the whole page scrolls- alternative fix #41

### LINKED ISSUE #41: Make the whole page scrolls

HooDeLin:
Fixes #41 
Staging site is [here](http://hoodelin.github.io/website)

This is an alternative fix. The previous fix is #104

### DIFF
--- a/scripts/common.js
+++ b/scripts/common.js
@@ -207,23 +207,13 @@ $(document).ready(function() {
         var week = id.substr(('header-content-week').length);
         addCollapseAndExpandButtonsForWeek('#' + id, 'content-week' + week);
     });
-    var bannerHeight = 25;
     var headerHeight = 40;
     var topMargin = 5;
     var topPadding = 5;
     $('#form-preferences').css('height', headerHeight);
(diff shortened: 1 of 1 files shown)


damithc:
I prefer this one. @wkurniawan07 ?

damithc:
BTW, it looks like we can stage only one version at a time.

wkurniawan07:
I don't have a particular preference. The trade-off for this version is the preference bar won't be so accessible when the user is so far down the page.
On an unrelated note, yes `gh-pages` can only stage one version of change at a time, unless the changes are independent (let's say this and a markdown conversion change).

damithc:
OK, let's go for this one then. I don't think the preference panel will be used often, so it makes sense not to allocate permanent screen real estate to it. If users don't like it, we can revert. @wkurniawan07 please proceed to review and merge.

Issue #81, opened 2015-12-10
Show a loading indicator in the handbook page


damithc:
Similar to #80, it takes some time to load the handbook page fully. May be we can show a loading indicator during that time?

PR #110, opened 2015-12-21
Show a loading indicator in the handbook page #81

crispyfridge:
@wkurniawan07 Hi, sorry to bother, but I need some advice here.
I added a loading gif which appears when the page is loading, and disappears after. However, the gif freezes while the page is loading. After doing a search it seems that this is caused by the heavy dom manipulation that goes on in the page while the script is loading. Following that, I tried to use webworker to create a separate background thread for the script so the image won't freeze, but the script contains jQuery, which webworker isn't compatible with. I'm not sure if converting the file to pure ajax would be a good idea, so I didn't do that.

I've also tried replacing the gif with a text, but the text always appears after the script has loaded, no matter what I did. I'm not sure what to do from this point on. :|

Thanks in advance.

### DIFF
--- a/contents/handbook.html
+++ b/contents/handbook.html
@@ -20,7 +20,7 @@
     </head>
     <body>
 
-
+<img id="load-spinner" src='../images/ajax-preload.gif' />
 <ul>
     <li><a href="#handbook-preliminaries">Preliminaries</a></li>
     <li><a href="#handbook-lectures">Lectures</a></li>
--- a/contents/handbook.html
+++ b/contents/handbook.html
@@ -20,7 +20,8 @@
     </head>
     <body>
 
-<p id="loading">Page Loading...</p>
+<div id="overlay"> </div>
+<div id="modal">Page Loading...</div>
 <ul>
     <li><a href="#handbook-preliminaries">Preliminaries</a></li>
     <li><a href="#handbook-lectures">Lectures</a></li>
--- a/contents/handbook.html
+++ b/contents/handbook.html
@@ -20,7 +20,7 @@
     </head>
     <body>
 
-<img id="load-spinner" src='../images/ajax-preload.gif' />
+<p id="loading">Page Loading...</p>
 <ul>
     <li><a href="#handbook-preliminaries">Preliminaries</a></li>
     <li><a href="#handbook-lectures">Lectures</a></li>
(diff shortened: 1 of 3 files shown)


damithc:
As a last resort, we can use a static image e.g. `loading ...`

wkurniawan07:
@crispyfridge hey, don't worry about it.
I don't think we can do much with the freezing gif, it's pretty much expected given that we have so much contents inside the handbook. And yeah I don't think writing a webworker/converting to pure ajax is worth the trouble.
What's wrong with using a good ol' text in place of the gif, though? I tried it and it worked just fine.

damithc:
Yes, I think text is a better idea. May be format it in a way it doesn't look like regular text.

crispyfridge:
The text doesn't appear while the content is loading on mobile, when tried on iPad and an android phone. Not sure if its just me though.

damithc:
The gif (un-animated) has a better success rate than text?

crispyfridge:
It always works on the desktop, but yes, the gif has a higher success rate on mobile devices. But I'm becoming suspicious of using my own mobile devices for tests, since they they are displaying different results compared to @wkurniawan07.
In any case, I've pushed the commit for the loading text, the preview is here:
http://crispyfridge.github.io/website/contents/handbook.html

damithc:
I don't like that the page moves slightly when the 'page loading...' disappears. Any way to avoid that? May be show the text as an overlay or something?

crispyfridge:
I've added an overlay with text over it, the preview is here:
http://crispyfridge.github.io/website/contents/handbook.html

damithc:
Nice. I wouldn't mind a smaller overlay though (not block the TOC too much). That way, users can read the TOC and locate the item they want while they wait for the page to load. But the current one is OK if a smaller one is hard to do.

crispyfridge:
I've reduced the size of the overlay, the preview is here:
http://crispyfridge.github.io/website/contents/handbook.html
Not sure if it is small enough though.

damithc:
The overlay looks fine :+1:

wkurniawan07:
Looks good to me as well :+1:
@crispyfridge, I think you have the permission to merge this yourself. Just use the Merge pull request button.

crispyfridge:
Ok got it thanks :)

crispyfridge:
Closed #110

PR #113, opened 2015-12-23
Fix size and position of 'Page Loading...' overlay in handbook page #112

### LINKED ISSUE #112: Fix size and position of 'Page Loading...' overlay in handbook page
and position near to top

crispyfridge:
Fixes #112 
Fixed size of overlay and moved its position closer to the top. 
The overlay is slightly off center in an iPhone 5, but is centered in other devices that I've tried.
Preview: http://crispyfridge.github.io/website/contents/handbook.html


damithc:
I tried the staging site from desktop. Didn't seem to work. The page was empty during the loading period.

crispyfridge:
Yes I have it frequently too. I just tested it on firefox on desktop it always works. It doesn't work on chrome and safari, but not sure what's causing it. Reverting it to the previous code gives me the same result, and removing the overlay completely gives me an empty page while loading as well. I can only guess something during the merge caused this, but I don't know what or how.

crispyfridge:
I've reverted the merge in my gh-pages branch and added the commit that I wanted. I don't have the blank page problem anymore. I'm not sure if the same works for you, you can try it here:
http://crispyfridge.github.io/website/contents/handbook.html

wkurniawan07:
This branch looks polluted with the changes from #108

crispyfridge:
Yeah, would it be better if I deleted this branch and branch from the master? That way it won't have the commits from #108

crispyfridge:
I'm going to close the pull request and create a new branch.
Closed #113

damithc:
As you are the only one using your fork, you can force-push to clean up past history. Learn things like interactive rebasing, squashing, cherry picking etc that can be useful when cleaning up a branch.

PR #114, opened 2015-12-24
Yellow highlight looks jarring on blue subheadings #23 

### LINKED ISSUE #23: Yellow highlight looks jarring on blue subheadings
1unnamed
An example above.

Something less saturated would work better.
(Or just **bold** it?)

ravic1995:
Fixes #23 
view : - http://ravic1995.github.io/website/

### DIFF
--- a/styles/common.css
+++ b/styles/common.css
@@ -330,13 +330,13 @@ table.bordered tr > th {
     font-weight: bold;
  }
 
-.highlighted,
-.highlighted-darktext {
+.highlighted{
     background-color: #ffff00;
 }
 
 .highlighted-darktext {
-    color: #2b2b2b;
+    color: #ffffff;
+    font-weight: bold;
 }
 
 .important {
--- a/styles/common.css
+++ b/styles/common.css
@@ -330,7 +330,8 @@ table.bordered tr > th {
     font-weight: bold;
  }
 
-.highlighted {
+.highlighted,
+.highlighted-darktext {
     background-color: #ffff00;
 }
 
@@ -338,11 +339,8 @@ table.bordered tr > th {
     color: #2b2b2b; 
 }
 
-li > .highlighted-darktext {
-    background-color: #ffff00;
-}
-
 .ui-accordion-header > .highlighted-darktext {
+    background-color: inherit;
     color: #ffffff;
     font-weight: bold;
 }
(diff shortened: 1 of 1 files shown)


wkurniawan07:
We can't do this. There's a legitimate use of yellow highlighting in week 5 contents and you're replacing it with white text (i.e invisible text).
If I may suggest, use this selector: `.ui-accordion-header > .highlighted-darktext`.

ravic1995:
@wkurniawan07 can you please elaborate a bit more. I am a bit unclear.

wkurniawan07:
This is part of the week 5 content in your staging version:
image
Your styling makes the text after "Important:" invisible.

The blue subheadings occur only in the accordion headers, which is why targeting the selector `.ui-accordion-header > .highlighted-darktext` should solve the problem without causing the above.

ravic1995:
@wkurniawan07 can you review again ?

styles/common.css
```diff
@@ -330,13 +330,21 @@ table.bordered tr > th {
...
 }
 
 .highlighted-darktext {
-    color: #2b2b2b;
+    color: #2b2b2b; 
+}
+
+li > .highlighted-darktext {
+    background-color: #ffff00;
+}
```
wkurniawan07: Why `li > .highlighted-darktext`?
Also, if this has the same style as `.highlighted` above, you can combine them.

ravic1995: So that the important content can be highlighted.  
If I don't include it the important content will still be "visible" (black font) but wont be highlighted as it currently is on the main site. 
screen shot 2015-12-31 at 11 24 29 am

wkurniawan07: You're making an assumption that everything else that have the class `highlighted-darktext` is part of `li`, which I don't think it's true.
Why not just leave it as before, i.e keep the `.highlighted, .highlighted-darktext { ... }` style, and if you want to suppress the `background-color` for `.ui-accordion-header > .highlighted-darktext`, use `background-color: inherit;` to override.

wkurniawan07:
The commits are good, but... you're making the PR from `master` branch? You're supposed to create a branch named something like `114-yellow-highlight-jarring` and make your PR from that branch.
Also, it might be too late to fix (but of course you're welcome to if you know how to), but you should use more informative commit messages next time.

wkurniawan07:
I haven't checked on this for a while. Since this branch is wrongly named, I'd have to close this PR.
@ravic1995 you can open another PR with the same change set with a correctly named branch, and it should be immediately good to merge.

PR #115, opened 2015-12-25
Remove 'revamped website' notice on top #96

### LINKED ISSUE #96: Remove 'revamped website' notice on top
This can be removed. However, we should still mention where to report bugs.
image

RyanAlexanderH:
Fixes #96 
Here is the [preview](http://ryanalexanderh.github.io/website/)

### DIFF
--- a/contents/headings.html
+++ b/contents/headings.html
@@ -33,7 +33,7 @@
                             <b>Announcements</b>
                         </a>
                         &middot;
-                        <a class="href" target="content" href="https://github.com/nus-cs2103/website">
+                        <a class="href" target="content" href="https://github.com/nus-cs2103/website/issues" target="_blank">
                             <b>Feedback</b>
                         </a>
                         &nbsp;&nbsp; 
--- a/contents/headings.html
+++ b/contents/headings.html
@@ -6,10 +6,6 @@
         <title>Headings</title>
     </head>
     <body leftmargin="5" topmargin="0">
-        <span class="highlighted">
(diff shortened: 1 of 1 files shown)


contents/headings.html
```diff
@@ -7,7 +7,7 @@
     </head>
     <body leftmargin="5" topmargin="0">
         <span class="highlighted">
-          This is the revamped version of the module website. Report bugs/suggestions 
+          Report bugs/suggestions 
           <a href="https://github.com/nus-cs2103/website" target="_blank">at our issue tracker</a>.
         </span>
```
damithc: Lets remove the entire `<span>` and add a 'Feedback' to the top navigation menu.

RyanAlexanderH:
Is it something like 'Feedback' on the top navigation menu linked to https://github.com/nus-cs2103/website?

damithc:
Yes, but you can link to the [issue tracker](https://github.com/nus-cs2103/website/issues) page instead.

wkurniawan07:
It seems like GitHub doesn't allow itself to be opened in a frame. In that case, open the link in a new tab instead (using `target="_blank"`).

RyanAlexanderH:
Updated the link to the [new one](https://github.com/nus-cs2103/website/issues) and added the new tab too

RyanAlexanderH:
Updated the link and now it should open the link in a new tab.

wkurniawan07:
Ok this looks good :+1:

PR #118, opened 2015-12-26
Previewable page fragments #89

### LINKED ISSUE #89: Make page fragments previewable

wzchua:
Fixes #89 
http://wzchua.github.io/website/

I used a global variable in the fragments to denote its week number. This is something tied to their class name and is needed for the js to apply the styles and effects correctly. I did not prettify the html as it made the diff hard to refer.


damithc:
Umm, does technique work for fragments of the handbook?

wzchua:
Yes, it would. I'll work on it.

wzchua:
The handbook fragments basically have the same css and style script the handbook.html has.

wkurniawan07:
Any demo so I can see if it works?

wzchua:
I added the gh-pages
Access the html directly base on the directory structure
For example:
http://wzchua.github.io/website/contents/week0.html

contents/handbook-appendixA.html
```diff
@@ -1,3 +1,25 @@
...
+        body,
+        table > tbody > tr > td,
+        table > tbody > tr > th {
+            font-size: 16px;
+        }
+    </style>
+</head>
+
+<body>
+    <div id="fragment">
```
damithc: You can keep this first div unindented. In any case, the indentation for html is 2 spaces, not 4.

contents/handbook-appendixA.html
```diff
@@ -1,3 +1,25 @@
...
+        body {
+            padding: 0px 10px;
+            max-width: 740px;
+        }
+        body,
+        table > tbody > tr > td,
+        table > tbody > tr > th {
+            font-size: 16px;
+        }
+    </style>
```
damithc: Any way to avoid having to copy paste this style segment in every file?

scripts/common.js
```diff
@@ -195,11 +184,26 @@ function loadContent(week) {
...
+            console.log("fragment week " + week + " load error");
+        } else {
+            console.log(status);
         }
-    });
+    })
 }
 
 $(document).ready(function() {
+    if(typeof _week_no !== 'undefined'){
```
damithc: Any reason why the variable `_week_no` starts with an underscore?

contents/handbook-appendixA.html
wkurniawan07: This is beyond the scope of this PR, but needs to be done. I'll do it right away. @wzchua you can remove all these inline styles, trust me it will be fine.

wzchua: Yes, we can group them into a file and load that html via js
http://stackoverflow.com/questions/18712338/make-header-and-footer-files-to-be-included-in-multiple-html-pages

scripts/common.js
wzchua: It is some other convention I've used for denoting private variables but I should not have done it here, I'll change that.

contents/handbook-appendixA.html
wzchua: It doesn't work well with js scripts

wzchua: The alternate approach to this is to have a preview.html that has the appropriate scripts and styles and will load the html based on an input

contents/handbook-appendixB.html
```diff
@@ -0,0 +1,15 @@
...
+<html>
+<head>
+  <meta content="text/html; charset=UTF-8" http-equiv="Content-Type" />
+  <link type="text/css" rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.11.3/jquery-ui.min.css" />
+  <link type="text/css" rel="stylesheet" href="../styles/common.css" />
+  <link type="text/css" rel="stylesheet" href="../styles/handbook.css" />
+</head>
+
+<body>
+  <div id="fragment">
```
damithc: This particular outermost div need not be indented. It looks odd when the div is indented but its content is not.

contents/handbook-appendixA.html
```diff
@@ -73,3 +84,7 @@
 Don't let that worry you. The final grade depends on what you know at the end, not what 
 you knew to begin with. The full 10% allocated to intermediate deliverables is within the reach 
 of everyone in the class irrespective of their prior knowledge.</p>
+
+  </div>
+</body>
+</html>
```
damithc: Apparently, html, head, body tags are optional. May be we should omit them in these fragments? @wkurniawan07 what do you think?

scripts/common.js
```diff
@@ -195,11 +184,26 @@ function loadContent(week) {
                     $('.' + type + '.content-week' + week).hide();
                 }
             });
+        } else if (status == 'error'){
+            console.log("fragment week " + week + " load error");
```
damithc: Unexpected errors should be shown in a more visible way so that they get noticed sooner?

scripts/common.js
```diff
@@ -195,11 +184,26 @@ function loadContent(week) {
...
+        } else {
+            console.log(status);
         }
-    });
+    })
 }
 
 $(document).ready(function() {
+    if(typeof week_no !== 'undefined'){
```
damithc: May be extract the condition into a local variable and give it an explanatory name?

contents/handbook-appendixA.html
wkurniawan07: I think the issue is on the repeated inline styling. Now that they're gone, it's no longer a problem. Calling external scripts and styles for each different HTML file is fine in web development if we're not using a templating engine (which we don't).

scripts/common.js
```diff
@@ -195,11 +184,26 @@ function loadContent(week) {
...
+        } else {
+            console.log(status);
         }
-    });
+    })
 }
 
 $(document).ready(function() {
+    if(typeof week_no !== 'undefined'){        
+        var week = week_no;
```
wkurniawan07: Rather than doing this, why not just `var week = 1/2/3/...;` in each of the page fragment?

contents/handbook-appendixA.html
wkurniawan07: Those three tags are optional in any HTML file. Let's just keep it to standardize.

scripts/common.js
wzchua: I need week_no to be global because I'm using the name js. var wouldn't give global scope and I can do something like this but I cannot use the name week. It seems to get overridden or reset.

damithc:
Any possibility of avoiding the extra div element? As that div simply wraps around the entire content of the body element, can't we use the `body` element instead?

wzchua:
The body element doesn't work. It gets filtered through the browser's internal process

damithc:
@wzchua can you syn this with the master branch and make it ready for review again?

wzchua:
I have synced but it looks someone's merge has gone wrong. handbook is now supposed to be a single file, right?

damithc:
No, only appendix B (policies) are supposed to be in a single file (previously each policy was in a separate file)

damithc:
GitHub says your branch has conflicts that needs to be solved.

wzchua:
okay, I see. Yea, I've only sync'ed the master

scripts/common.js
wzchua: I'm not sure if alert is the best way to inform. the 15 fragments are loaded consecutively, which can result in 15 alerts

wzchua:
I've sync'ed the issue branch.

acjh:
Fixed in #209

PR #122, opened 2015-12-27
Scroll section into view when expanded #42

jkt001:
This is an on-going attempt at fixing #42 

Staged version is <a href = "http://jkt001.github.io/website/">here</a>

Currently it supports scrolling to weekly sections, but not into sections inside the weekly schedule. Is this the expected way (since people would read the sections in a given week anyway, after scrolling directly to that week)?

PR #123, opened 2015-12-28
Scroll section into view when expanded #42

### LINKED ISSUE #42: Scroll section into view when expanded
When user expands the sections that are positioned near the bottom of page, the expanded content are out of view:
screenshot 2015-10-09 13 54 05

Perhaps it'll be better to auto-scroll the content into view after clicking:
screenshot 2015-10-09 13 54 15

jkt001:
Fixes #42 and supersedes closed PR #122 

Staged version can be viewed <a href = "http://jkt001.github.io/website">here</a>.

Currently, the fix is to scroll to a "week section" and not to inner sections since I assume that people would most likely scroll into a specific week and read the contents inside. 

If this is not the expected behaviour, I would update this PR in subsequent commits.

### DIFF
--- a/scripts/common.js
+++ b/scripts/common.js
@@ -199,6 +199,15 @@ function loadContent(week) {
     });
 }
 
+function addAutoScrollToClickedSection() {
+ $(".buttoned").click(function (event){
+   var scroll_target = "#" + event.target.id;
+   $('html, body').animate({
+     scrollTop: $(scroll_target).offset().top
+ }, 1000);
+});
+}
+
 $(document).ready(function() {
 
     makeAccordion('.weeklyschedule');
@@ -220,6 +229,7 @@ $(document).ready(function() {
         loadContent(week);
     }
 
+    addAutoScrollToClickedSection();
     addCollapseAndExpandButtonsForAllContents("#form-preferences");
 
     // toggles showing/hiding certain sections according to the preferences checkbox
--- a/scripts/common.js
+++ b/scripts/common.js
@@ -202,7 +202,6 @@ function loadContent(week) {
 function addAutoScrollToClickedWeekHeader() {
     $('.buttoned').click(function(event) {
         var scrollTarget = '#' + event.target.id;
-        console.log(scrollTarget);
         $('html, body').animate({
             scrollTop: $(scrollTarget).offset().top 
         }, 100);
--- a/scripts/common.js
+++ b/scripts/common.js
@@ -199,12 +199,12 @@ function loadContent(week) {
     });
 }
 
-function addAutoScrollToClickedSection() {
-    $(".buttoned").click(function(event) {
-        var scroll_target = "#" + event.target.id;
+function addAutoScrollToClickedWeekHeader() {
+    $('.buttoned').click(function(event) {
+        var scrollTarget = '#' + event.target.id;
         $('html, body').animate({
-            scrollTop: $(scroll_target).offset().top 
-        }, 1000);
+            scrollTop: $(scrollTarget).offset().top 
+        }, 0);
     });
 }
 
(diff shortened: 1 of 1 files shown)


damithc:
@wkurniawan07 for your review.

jkt001:
I have merged master into this branch to reflect the most up-to-date version of the website as in the master branch. Staged version is still the same as above.

scripts/common.js
```diff
@@ -199,6 +199,15 @@ function loadContent(week) {
...
 
+function addAutoScrollToClickedSection() {
+ $(".buttoned").click(function (event){
+   var scroll_target = "#" + event.target.id;
+   $('html, body').animate({
+     scrollTop: $(scroll_target).offset().top
+ }, 1000);
+});
+}
+
```
wkurniawan07: Some coding standard violations here:

1. Strings should be wrapped with `'` instead of `"`
2. Indentations are inconsistent. Whether you want it to be 2 or 4 spaces is up to you (try to follow the surroundings, though), but be consistent.
3. We use `camelCase` for variables, not `snake_case`.
4. `click(function(event) {`
5. And maybe rename the function name as well. This auto-scroll is equipped to `.buttoned` class, which from what I remember is just all the week headers. "Section" is too ambiguous (and somewhat inaccurate) here.

And regarding the function itself: the value `1000` refers to the time taken for the scrolling to finish. 1 full second might be too long? Try to adjust it, or maybe just omit it altogether (using the value `0`), see which one looks better.

damithc: Indentation should be 2 spaces for HTML and 4 spaces for other languages.

jkt001:
I have fixed the PR accordingly. Staged version can be viewed at the same location <a href = "http://jkt001.github.io/website/">here</a>.

For the time to scroll, I've tried 0 and it looks rather confusing to the user since they automatically move to that section (just my opinion). I tried using 100ms and you all can see how it looks.

damithc:
I tried the staged version just now. It doesn't seem to scroll to view. May be github.io takes some time to update?

jkt001:
@damithc 

Thanks for the heads up. You could now try again from the staged version <a href = "http://jkt001.github.io/website/">here</a>.

It seems that there was a problem when I used target instead of currentTarget.

I've tried it and it seems to be fixed. Let me know if there's anymore problem.

damithc:
Still the same problem. Does it work for you when you try to staging site?

jkt001:
I see. It worked for me when I tried the staging site but then I just observed another problem (which is the one you mentioned).

The problem seems to be tied with the time to scroll. I changed it to 500 ms and now it seems that the problem is gone. Previously, what I noticed was it did not scroll to the clicked section. That problem was solved with the previous commit ef07f9f. However, what I now noticed after trying it again was that it scrolls, just not to the clicked section completely. This is now solved and it should work just fine. 

Staged version can be viewed <a href = "http://jkt001.github.io/website/">here</a>. It should work now, tested in Firefox and Chrome.

damithc:
I see. It is odd how a change in the time has this kind of effect though. :-p

jkt001:
Yeah it seems odd. I previously suspected that the function was not called properly, hence I tried logging to see if the function is actually called when users click on the weekly headers. Then I found out that the function <b>is</b> called. 

After tweaking with the time to scroll, I found out that setting it to 500 ms solves the problem. Initially, I set it to 1000 ms and it seemed a bit slow as @wkurniawan07 has mentioned. 

I believe that the current staged version works OK and 500 ms is a reasonable time to scroll. Setting it less might be too "direct" for users and the current one nicely simulates the scrolling process without taking too much time. What do you think?

wkurniawan07:
Ok this is good to go :+1:

PR #124, opened 2015-12-28
Schedule page: generate start date of weeks automatically #119

### LINKED ISSUE #119: Schedule page: generate start date of weeks automatically
image
Currently these are hard coded. They can be generated using JS. That way, only the start date needs to be changed when reusing the website in subsequent semesters.

HooDeLin:
Fixes #119


scripts/common.js
```diff
@@ -199,8 +199,25 @@ function loadContent(week) {
...
+    for (var week = 1; week <= 13; week ++) {
+        if (week == 7) { //recess week
+            moduleStartDate.setDate(moduleStartDate.getDate() + 7);
+        }
+        moduleStartDate.setDate(moduleStartDate.getDate() + 7);
+        $('#header-content-week' + week).append('<span>Week ' + week + ' [' + month[moduleStartDate.getMonth()] + ' ' + moduleStartDate.getDate() + ']</span>');
+    }
+    $('#header-content-week14').append('<span>Reading Week</span>');
+}
+
```
damithc: Just wondering, is it possible to make this more general by having a function `getDate(week, day)`? e.g. `getDate(1,Tue)` returns the date week 1 Tuesday. That way, we can use the generation mechanism anywhere in the website we want. Right now, it is tightly coupled to the schedule page.

HooDeLin: I think it is only possible if we have a way to retrieve the starting date of a semester since the starting week differs every semesters. However, I can provide a `getDate(week, day, startingDate)` function where `startingDate` is the first day of the semester to reduce coupling to some extent.

damithc: May be starting day can be defined as a global constant?

HooDeLin: Yes, that is possible. If that is the case, the dates can be updated for any semesters by changing the starting day.

damithc: Ok, go ahead and implement that approach.

HooDeLin:
The date retrieving has been refactored out to  `getDate(week, day)` function by setting the start date of a semester as a global variable.

scripts/common.js
```diff
@@ -199,8 +199,31 @@ function loadContent(week) {
...
+function getDate(week, day) {
+    var date = new Date();
+    if (week >= 7) {
+        date.setTime(window.moduleStartDate.getTime() + (week * 7 * 24 * 60 * 60 * 1000) + ((day - 1) * 24 * 60 * 60 * 1000));
+    } else {
+        date.setTime(window.moduleStartDate.getTime() + ((week - 1) * 7 * 24 * 60 * 60 * 1000) + ((day - 1) * 24 * 60 * 60 * 1000));
+    }
+    return date;
+}
+
```
damithc: We are still injecting dates to the html page from outside. What I'm looking for is a way to embed some markers in the html to indicate where I want the dates. For example, if I have a deadline on week 5 Thursday, I should be able to add some marker in the relevant place of week 7 and the correct date will be displayed when the page is loaded. Is that possible?

HooDeLin: Do you mean that for any dates in the document, we just put markers there and let the js figure out the date?

For example, we can specify a deadline on week 5 Thursday by just having `<div>5,4</div>` in the document. Then the correct date will be displayed when the page is loaded. Is it what you are looking for?

damithc: Yes, something like that. I'm not sure how hard it is to do using plain JS.
@zeulb do you think Angular can be useful here?

zeulb: Yes, it is possible. But I think we can also use jQuery here. Let say we have something like this `<span class="dateMarker" week="5" day="thursday"></span>` in our html file, we can tell jQuery to select elements with class `dateMarker`, then for each of those elements, jQuery will do the calculation and modify the element to something like this `<span class="dateMarker" week="5" day="thursday">Dec 13</span>`.

HooDeLin: Yes, I believe that jQuery can handle it.

HooDeLin:
@damithc 
I have done an alternative version where dates are wrapped in `<span class="dateMarker" week="5" day="1"></span>`. In this way, common.js only generates dates based on the html file instead of injecting them directly. Is this what you are looking for?

This is just an experimental version. It is still insufficient to corporate other date formats found in the html file. Right now, it is working fine for the time in the schedule page.

damithc:
Yes, this is fine. No need to accommodate other date formats for now. We can use the same date format in other places too. e.g. `Fri [<span class="dateMarker" week="5" day="1"></span>] 23.59pm`
May be you can add dates in other places of the schedule page where the date can be useful (e.g. deadlines)

HooDeLin:
I have added dateMarkers into deadline in the schedule content
This is the [staging site](http://hoodelin.github.io/website/)

scripts/common.js
```diff
@@ -199,8 +200,32 @@ function loadContent(week) {
...
 $(document).ready(function() {
 
+    //Note: Javascript's month are zero-indexed
+    window.moduleStartDate = new Date(2015, 7, 10);
```
damithc: Is it possible to define this as a constant, in a more noticeable place e.g. at the top of this file?

damithc:
@wkurniawan07 for review please

scripts/common.js
HooDeLin: We can wrap the initialisation into a function and place that function at the top of the file.
We can also shift the whole `$(document).ready(function() { ... });` to the start of the file.
I think the first approach is better but I am not sure if there is any other possible ways though.

HooDeLin:
I have refactored the initialisation of the date variables to the top of the file

scripts/common.js
```diff
@@ -199,8 +207,30 @@ function loadContent(week) {
...
+        var week = parseInt($(this).attr('week'));
+        var day = parseInt($(this).attr('day'));
+        var date = getDate(week, day);
+        $(this).html(window.month[date.getMonth()] + ' ' + date.getDate());
+    });
+}
+
+function getDate(week, day) {
+    var date = new Date();
+    if (week >= 7) {
```
wkurniawan07: It is unclear why is a distinction made for `week >= 7`, i.e there's no guarantee that the future developers of this can tell straightaway that this is because of recess week. Add a comment to explain on this.

scripts/common.js
```diff
@@ -199,8 +207,30 @@ function loadContent(week) {
...
+        date.setTime(window.moduleStartDate.getTime() + (week * 7 * 24 * 60 * 60 * 1000) + ((day - 1) * 24 * 60 * 60 * 1000));
+    } else {
+        date.setTime(window.moduleStartDate.getTime() + ((week - 1) * 7 * 24 * 60 * 60 * 1000) + ((day - 1) * 24 * 60 * 60 * 1000));
```
wkurniawan07: `* 24 * 60 * 60 * 1000` is repeated four times. See if you can extract it, as a helper function or as local variables should be good enough as I don't think it's used anywhere else.

scripts/common.js
```diff
@@ -219,7 +249,7 @@ $(document).ready(function() {
         $('#content-week' + week).html('<img height="40" width="40" class="margin-center-horizontal" src="/images/ajax-preload.gif"/>');
         loadContent(week);
     }
-
+
```
wkurniawan07: Undo this?

scripts/common.js
```diff
@@ -199,8 +207,30 @@ function loadContent(week) {
     });
 }
 
+function generateDates() {
+    $('.dateMarker').each(function() {
```
wkurniawan07: `date-marker`? We use `spinal-case` (or whatever this is called) for CSS classes and ids.

wkurniawan07:
Aside from the small improvements I suggested, this looks good.

HooDeLin:
The improvements are done

scripts/common.js
```diff
@@ -199,8 +207,32 @@ function loadContent(week) {
...
+}
+
+function getDate(week, day) {
+    var date = new Date();
+    var dayToMiliSec = 24 * 60 * 60 * 1000;
+    // There needs to be an extra week after week 7 due to recess week
+    if (week >= 7) {
+        date.setTime(window.moduleStartDate.getTime() + (week * 7 + day - 1 ) * dayToMiliSec);
+    } else {
+        date.setTime(window.moduleStartDate.getTime() + ((week - 1) * 7 + day - 1) * dayToMiliSec);
```
wkurniawan07: You can try this to make it even less repetitive:
```javascript
if (week < 7) {
    week -= 1; // There needs to be an extra week after week 7 due to recess week
}
date.setTime(...)
```
Then you can also extract the `week * 7 + day - 1` to an explanatory variable, maybe something like `numOfDaysSinceStartDate`.
And shouldn't it be `after week 6` or `from week 7 onwards`? If I remember correctly recess week is between weeks 6 and 7.

HooDeLin:
The changes are done to reduce repetition

scripts/common.js
```diff
@@ -199,8 +207,31 @@ function loadContent(week) {
...
+        $(this).html(window.month[date.getMonth()] + ' ' + date.getDate());
+    });
+}
+
+function getDate(week, day) {
+    var date = new Date();
+    var dayToMiliSec = 24 * 60 * 60 * 1000;
+    if (week < 7) { // There needs to be an extra week from week 7 onwards due to recess week
+        week -= 1;
+    }
```
damithc: This code does not follow intuition, although it works.
A more intuitive way to implement the logic is (if JS syntax allows it):
```
isAfterRecessWeek = (week > 6);
weeksPassed = (week -1) + (isAfterRecessWeek ? 1 : 0);
daysPassed = weeksPassed * 7 + (day - 1);
```
This code, although a bit long winded, follows the natural logic of the situation (i.e. normally 'weeks passed' is one less than the week number, but have to add 1 if it is after recess week) and is self-explanatory.

HooDeLin: Ok, I'll work on it.

HooDeLin:
The function is more intuitive now

scripts/common.js
```diff
@@ -199,8 +207,31 @@ function loadContent(week) {
...
+        var week = parseInt($(this).attr('week'));
+        var day = parseInt($(this).attr('day'));
+        var date = getDate(week, day);
+        $(this).html(window.month[date.getMonth()] + ' ' + date.getDate());
+    });
+}
+
+function getDate(week, day) {
+    var date = new Date();
+    var dayToMiliSec = 24 * 60 * 60 * 1000;
```
damithc: variables should be named to sound like 'things', not 'actions'. e.g. `milliSecsPerDay`

scripts/common.js
```diff
@@ -199,8 +207,31 @@ function loadContent(week) {
...
+    var isAfterRecessWeek = week > 6;
+    // Add an extra week if the week number is after recess week
+    var weeksPassed = week - 1 + isAfterRecessWeek;
```
damithc: Better. But if the code is self-explanatory, you don't need the comment. The reason you need the comment is because the code is not self-explanatory enough. The explanation was lost when you reduced the logic into a formula. The formula does not say 'normally it is one less than `week`, but if we are past recess week we need to add one more". Instead, the formula says 'take the `week`, deduct 1, add condition `isAfterRecessWeek` and the result is the number of weeks passed'.

scripts/common.js
```diff
@@ -199,8 +207,31 @@ function loadContent(week) {
...
+    var daysPassed = weeksPassed * 7 + day - 1;
+    date.setTime(window.moduleStartDate.getTime() + daysPassed * dayToMiliSec);
+    return date;
+}
```
damithc: Side question: Can I get a day of recess week by providing something like week 6, day 9 ?

scripts/common.js
```diff
@@ -199,8 +207,31 @@ function loadContent(week) {
...
+function generateDates() {
+    $('.date-marker').each(function() {
+        var week = parseInt($(this).attr('week'));
+        var day = parseInt($(this).attr('day'));
+        var date = getDate(week, day);
+        $(this).html(window.month[date.getMonth()] + ' ' + date.getDate());
+    });
+}
+
+function getDate(week, day) {
```
damithc: Add a header comment for this function?

scripts/common.js
```diff
@@ -1,3 +1,10 @@
+function initDateVariables() {
+    // Change this to update the dates
+    // Note: Javascript's month are zero-indexed
+    window.moduleStartDate = new Date(2015, 7, 10);
+    window.month = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
+}
```
damithc: Why not declare a global variable instead? e.g.
`var MODULE_START_DATE = "2015.8.10";`
@wkurniawan07 Any downside to that?

scripts/common.js
HooDeLin: Yes, you can

damithc: OK, that's good. In that case, we don't have to add extra code to handle dates in the recess week.

scripts/common.js
HooDeLin: I am not sure if there is any serious downside. However, if we use `var MODULE_START_DATE = '2015.8.10'`, we will not be able to wrap it inside a function because it will be a local variable.

HooDeLin:
`dayToMiliSec` has been renamed to `miliSecsPerDay`
Header comments are added for generateDates() function

scripts/common.js
```diff
@@ -199,8 +207,35 @@ function loadContent(week) {
     });
 }
 
+/**
+ * generateDates() iterates through 'date-marker' span class and generate the corresponding dates.
```
damithc: No need to mention the name of the function. `Iterates through ...`

damithc: generate --> generates

scripts/common.js
```diff
@@ -199,8 +207,35 @@ function loadContent(week) {
...
+ * To specify a date, add '<span class="date-marker" week="1" day="1"></span>' in the html file.
```
damithc: You did not specify the meaning of `week` and `day` parameters. 
Also need to mention how to set the start date.

scripts/common.js
```diff
@@ -199,8 +207,35 @@ function loadContent(week) {
...
+        var week = parseInt($(this).attr('week'));
+        var day = parseInt($(this).attr('day'));
+        var date = getDate(week, day);
+        $(this).html(window.month[date.getMonth()] + ' ' + date.getDate());
+    });
+}
+
+function getDate(week, day) {
+    var date = new Date();
+    var miliSecsPerDay = 24 * 60 * 60 * 1000;
```
damithc: mili --> milli ?

HooDeLin:
Comments are updated

scripts/common.js
```diff
@@ -199,8 +208,38 @@ function loadContent(week) {
...
+        var week = parseInt($(this).attr('week'));
+        var day = parseInt($(this).attr('day'));
+        var date = getDate(week, day);
+        $(this).html(window.month[date.getMonth()] + ' ' + date.getDate());
+    });
+}
+
+function getDate(week, day) {
+    var date = new Date();
+    var milliSecsPerDay = 24 * 60 * 60 * 1000;
```
damithc: In Java, we would name this as MILLI_SECS_PER_DAY, but not sure what is the convention for JS. I'll leave it to @wkurniawan07 to decide.

scripts/common.js
```diff
@@ -1,3 +1,11 @@
+function initDateVariables() {
+    // Change this to set the module start date
+    // Format: Date(Year, Month, Day)
+    // Note: Javascript's month are zero-indexed
+    window.moduleStartDate = new Date(2015, 7, 10);
+    window.month = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
+}
+
```
damithc: It makes sense to define constants at the top of the file to make them easy to find/modify. But I'm not sure if extracting this function and putting it at the top of the file has the same benefit. If there is no way to define it as a constant (or to look like a constant e.g. `var MODULE_START_DATE = "2015.08.10"`), then may be we don't need to extract this function here? In any case, there is no need to put `window.month` here. @wkurniawan07 what do you think?

scripts/common.js
wkurniawan07: It's the same for js, so `MILLI_SECS_PER_DAY` should be good.

scripts/common.js
wkurniawan07: `window.*` is usually a technique to define a global variable in run time, but in this case we don't need such thing so both variables can be global without the need of this init function nor `window.*`. `var MODULE_START_DATE = ...` and `var MONTHS = ...` should do good enough.

damithc:
This branch is running into some merge conflicts due to recent updates to master. Can you sync this branch with upstream repo?

HooDeLin:
The branch is synced and the required changes are made

scripts/common.js
```diff
@@ -1,3 +1,9 @@
+// Change this to set the module start date
+// Format: Date(Year, Month, Day)
+// Note: Javascript's month are zero-indexed
+var MODULE_START_DATE = new Date(2015, 7, 10);
```
damithc: May be no need to bother the user with JS zero-index issue? Let the user set it as a string in the natural format e.g. `2015.08.10` and convert to a proper JS Date internally in your code.

HooDeLin:
The user can now enter the dates in a natural format. Format: `'Year.Month.Date'`

scripts/common.js
```diff
@@ -210,6 +252,7 @@ function addAutoScrollToClickedWeekHeader() {
 
 $(document).ready(function() {
 
+    generateDates();
```
wkurniawan07: Can we just call `generateDates()` once? From what I can tell, if this is called after all the contents are loaded (that is, after looping the number of weeks and loading each of the content), the dates will still be generated correctly.

wkurniawan07:
One small improvement and it's good to go. Well done! :+1:

scripts/common.js
HooDeLin: Yes, we can call this once. However, we are unable to call it after all contents are loaded. It somehow does not recognize it. We need to call it inside `loadContent(week)` and the `generateDates()` will be called every single loop.

HooDeLin:
The improvements are made

scripts/common.js
```diff
@@ -1,3 +1,17 @@
+// Change this to set the module start date
+// Format: 'Year.Month.Date'
+// E.g. '2015.8.10'
+var MODULE_START_DATE = setModuleStartDate('2015.8.10');
+var MONTH = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
```
damithc: Shouldn't this be MONTHS (i.e. plural)? Can be more specific too MONTH_NAMES_SHORT_FORM

HooDeLin:
The MONTH variable has been renamed to MONTH_NAMES_SHORT_FORM

wkurniawan07:
This is good to go :+1:

PR #128, opened 2015-12-29
Remove headings.js and its reference #121

### LINKED ISSUE #121: Remove headings.js and its reference
The file headings.js is not inside the contents folder, but is located in the scripts folder (which is not inside the contents folder). 

screen shot 2015-12-28 at 04 56 25

the file headings.html is inside the contents folder, hence the current way of calling headings.js will always result in a file not found error.

sihaoo:
Fixes #121
Updated headings.js to point to the correct javascript directory


wkurniawan07:
Actually, this is a page function that is no longer needed. You can remove the entire `headings.js` file and its reference in `headings.html`. I'll change the issue and PR titles for you.

damithc:
I think @wkurniawan07 wants you to remove the file as well.

wkurniawan07:
Ok, this is good :+1:

PR #132, opened 2015-12-30
Support proper browsing on mobile device #27

### LINKED ISSUE #27: Support proper browsing on mobile device
There's no mobile view, is there?
For example, on Firefox on Android, it shows the whole page without any attempt to resize anything.
As a result, everything is tiny and when zoomed in, you have to manually scroll horizontally to read what is there.
The right-side buttons are pretty difficult to tap too.

vhazali:
Fixes #27 

Not sure if this fulfils the requirement of "proper browsing". There's some quirks like images being squeezed very small and placed beside texts (for lecture tab under schedule). Still working on a way to change that without removing it from its table structure.


damithc:
Preview link seems to be incorrect.

vhazali:
Sorry; edited the link.

styles/common.css
```diff
@@ -397,4 +397,15 @@ table.bordered tr > th {
...
+
+img {
+	max-width: 100%;
+	height: auto;
+}
+
+@media screen and (max-device-width: 1000px) {
+	body{
+		max-width: 100%;
+	}
```
wkurniawan07: Use 4 spaces, not tabs, for indentation.

styles/handbook.css
```diff
@@ -1,3 +1,13 @@
+body {
+	padding: 0px 10px;
+	max-width: 740px;
+}
+body,
+table > tbody > tr > td,
+table > tbody > tr > th {
+	font-size: 16px;
+}
+
```
wkurniawan07: You can remove everything above this, I've tidied this part up in #130.

wkurniawan07:
You might want to put the meta viewport in these files as well:

1. schedule.html
2. coding-standards-cpp.html
3. coding-standards-java.html
4. pointsToPonderAboutProductDesign.html
5. testingFaq.html

vhazali:
Done the changes requested! I removed the styling from handbook.css but not sure if I should return them to the inline style in handbook.html?

wkurniawan07:
You can just sync with the current master, it will be automatically updated. If `handbook.html` conflicts during sync (which shouldn't, I think), then you should revert it first.

wkurniawan07:
The code looks good for now. Ping me again after syncing and I'll have one (hopefully) last review.

vhazali:
Alright, merged the latest master. @wkurniawan07

wkurniawan07:
Minor display tweaks:

1. The first line of code example in `handbook-ce1.html` is very long and will very likely overflow in smaller screen size. You can move everything after "(OR ..." to a new line, just a single line break (don't leave an empty line in between) will do.
2. In `handbook-textBooks.html`, the supplementary books image will be out of proportion because the height is also specified. Remove the height attribute.

After that this should be good to go. Great job! :+1:

vhazali:
Done! 

Thanks

contents/handbook-ce1.html
```diff
@@ -4,7 +4,8 @@
 </p>
 
 <pre class="cliprint">
-<code class="clisystem">c:></code> <code class="cliuser">TextBuddy.exe mytextfile.txt</code> (OR <code class="clisystem">c:></code> <code class="cliuser">java TextBuddy mytextfile.txt</code>)
+<code class="clisystem">c:></code> <code class="cliuser">TextBuddy.exe mytextfile.txt</code><br>
```
wkurniawan07: We don't need the `<br>` here. It defeats the purpose of not having any empty line in between these two lines.

contents/handbook-textBooks.html
```diff
@@ -5,7 +5,7 @@
 <p>
     Here are some relevant books you can read to supplement module content (optional reading).
 </p>
-<img src="../images/supplementary books.png" style="width: 716px; height: 211px;">
+<img src="../images/supplementary books.png" style="width: 716px">
```
wkurniawan07: I think you accidentally removed the first `;`.

index.html
```diff
@@ -3,6 +3,7 @@
     <head>
         <meta content="text/html; charset=UTF-8" http-equiv="Content-Type" />
         <meta http-equiv="Content-Language" content="en-us">
+        <meta name="viewport" content="width=device-width, initial scale=1" />
```
wkurniawan07: I cross-checked with online examples:
```html
<meta name="viewport" content="width=device-width, initial-scale=1.0">
```
The 1.0 and 1 difference might be fine, but the initial scale without hyphen, does it have a potential to break?

vhazali:
Corrected the mistakes. Thanks for pointing out the missing ```-```

wkurniawan07:
This is good now. There's actually more images that have pre-set height attributes and need removal, but you've done enough here. Great job! :+1:

vhazali:
@wkurniawan07 so regarding all those images with the height attributes, I'll just create an issue and fix them?

Issue #117, opened 2015-12-26
Cut down loading time of handbook page


damithc:
May be some form of incremental or on-demand loading instead of trying to load everything at once and freezing the page until everything is loaded?

jkt001:
I have a suggestion on this issue:

1. We "split" the sections of the handbook according to the bullet points (i.e. Preliminaries, Lectures, Project, etc.) into a separate HTML file. Each of that HTML file will then load all sub-sections accordingly.

2. In order to split according to (1), we can combine multiple short sections (such as Programming Languages / Textbook) into one section and perhaps split up the Project section to distinguish the CEs and the main product itself.

3. The main handbook.html page will contain links to each section, which will then load the appropriate contents. The main page can also have some nice cover / title to specifically say that it is the handbook (the current one does not have it).

This way, when students view the main handbook page, they will see links to each section. The page itself would only load the respective links to each section. Hence, unnecessary information (such as the project information when students only want to view exam information) is not loaded since the student would only click the link to the "Exam" subsection and that would only load information regarding the exam.

If my suggestion is ok, I will work on it as soon as possible. Any subsequent changes can be made after a preliminary version is done.

damithc:
The reason why we load everything is because it allows users to find things using Ctrl+F. We may not need that anymore when the site-wide search is implemented (currently in progress). At that point, we can change the handbook page to a more load-on-demand model.
The reason why we have split it into so many small parts is because some of those parts are embedded in other pages. When #32 is done, we can combine some of those fragments into bigger pages (e.g. FAQ, Policies). That will reduce the number of calls to the server.
I suggest we hold off this issue for now, until the site-wide search and #32 are done.

damithc:
@jkt001 You can give #32 a try if you are interested :-)

PR #133, opened 2015-12-30
Add a way to load a specific part of a page into an embedded link #32

### LINKED ISSUE #32: Add a way to load a specific part of a page into an embedded link
Currently, embedded links loads the entire page into the page (loading sections of the handbook in the schedule page). In some cases, we might want to load only one segment of the page.

jkt001:
Fixes #32

Staged version can be viewed <a href = "http://jkt001.github.io/website">here</a>.

The staged version is combined with PR #123 as I could not stage two different versions at the same time without combining it.

This version basically has no visual changes to the previous one. It just changes the way of loading the handbook page. The pages in the handbook is not merged yet (i.e. into one FAQ html page, one policy page etc) to avoid possible conflict with #118). It is easy (albeit tedious) to merge the pages together later, with the need to change only the file name that it is loaded (since the div id will stay the same). This can be done in a (new) separate issue or done in this same PR.

Edit: it is better to do it in another issue since this issue is to add a way of loading a specific part of a page into an embedded link, which has been done. Just my opinion :) .


contents/handbook.html
```diff
@@ -247,73 +247,78 @@
...
+        'handbook-faq-noLaptop handbook-faq-noLaptop',
+        'handbook-faq-narrowScope handbook-faq-narrowScope',
+        'handbook-faq-vagueRequirements handbook-faq-vagueRequirements',
+        'handbook-faq-favoriteTool handbook-faq-favoriteTool',
+        'handbook-faq-manySubmissions handbook-faq-manySubmissions',
+        'handbook-faq-intermediateMarks handbook-faq-intermediateMarks',
+        'handbook-faq-cs2101Differences handbook-faq-cs2101Differences',
+        'handbook-appendixD handbook-appendixD',
+        'handbook-appendixE handbook-appendixE',
+        'handbook-appendixF handbook-appendixF'];
```
wkurniawan07: If they're always the same, why don't just specify it once?

contents/week1.html
```diff
@@ -72,7 +72,7 @@ <h3 id="things-to-do-content-week1" class="buttoned"><span>Things to do</span></
         <div class="things-to-do things-to-do-week1">
             <h3>Attend Lecture 1 <span class="highlighted-darktext">(compulsory)</span></h3>
             <div>
-                <span onclick="getContentUsingAjax('handbook-lectures', '#embedded-week1-lectures');"><span class="embedded-link">More about lectures</span></span>
+                <span onclick="getContentUsingAjax('handbook-lectures', '#embedded-week1-lectures', 'lectures');"><span class="embedded-link">More about lectures</span></span>
```
wkurniawan07: Same for these. It looks like the 3rd param is taken from the 1st param with `handbook-` truncated from it. Let's not repeat stuffs unnecessarily.

contents/handbook.html
jkt001: they would change in the future if the FAQ or policy file is combined into one.

The left one specifies the file that we want to extract information from.

So, for example, if we have a single handbook-faq.html file with multiple sections (these sections are the same as in the links in the top of the handbook page). 

The sections on the handbook-faq.html would then be loaded into those sections. 

I just realised that we need to specify one more section, i.e. the section from the original file that needs to be loaded into the specified section of the page, e.g.

handbook faq-noLaptop handbook-faq-noLaptop

means:

the content is located in handbook.html in the section with id faq-noLaptop and it is to be loaded into the section with id handbook-faq-noLaptop in the handbook.html file that is displayed to the user.

If this is not required as of now (since the files has not been merged into one), I can change it back to the previous version and have another issue to merge the faq / policy / etc. into one file. I believe this can be done on a separate issue (i.e. to merge the files into one).

contents/week1.html
jkt001: this is necessary since we need the pullContent and getContentUsingAjax to be able to load a specific section of the specified html file.

Once we have a policy.html file for example, the function call would look something like this:

getContentUsingAjax(policy, '#some-name-here', 'policy-reuse')

which gets the section with id 'policy-reuse' from the file policy.html.

This cannot be done in a separate issue since this is to solve the current issue of having a way to extract a specific part of a page into an embedded link.

contents/handbook.html
wkurniawan07: I don't see much use of this. If we're still loading the whole of handbook by default, then even if we combine some sections into one we can simple load the whole combined sections. Until we're doing on-demand load (#117), of course, but hey let's not jump the queue, one issue at a time.

contents/week1.html
wkurniawan07: Ok, got it. This one is necessary alright.

scripts/common.js
```diff
@@ -6,25 +6,23 @@ function makeAccordion(elementSelector) {
     });
 }
 
-function getContentUsingAjax(section, elementSelector) {
-    pullContent(section, elementSelector, 'Exract from handbook');
+function getContentUsingAjax(section, elementSelector, sectionName) {
+    pullContent(section, elementSelector, 'Exract from handbook', sectionName);
```
wkurniawan07: Rename `section -> fileName` to reflect its usage now. Similarly in `pullContent`.

contents/handbook.html
jkt001: Yes I agree with this. For now, I'll revert this back to the original version. We can then have another issue specifically to combine the faq and policy files into one html file (and possibly other files as well?). This seems like a big task worthy of another issue.

scripts/common.js
jkt001: hmm... is it necessary to change the name? 

My reasons not to:
<ol>
<li> It actually <b>does reflect the section name</b>, i.e. the section name is, for example, <code>policy-reuse</code> while its filename is <code>handbook-policy-reuse</code>, hence it does not actually represent the file name, but the section name.</li>

<li>In the future, the name needs to be changed again. Since the current one also calls in the sectionName, we can just let it be.</li>
</ol>

This is just my opinion though.

damithc:
Good progress on this issue @jkt001 :+1: 
Can we make the 3rd parameter optional? i.e. the original two parameter method will load the entire page content. A new method that takes the third parameter will load the specified div only. 
Once you have done the above, there is no need to change all the existing method calls and page fragments. So you can revert those changes.
After that, combine fragments of one sections. e.g. Policies

jkt001:
@damithc I have made the third parameter optional via 40a296a.

Staged version (with the latest master) is available <a href = "http://jkt001.github.io/website/">here</a>.

For now, it's just a matter of combining the files into one (i.e. FAQ, Policy).

Do you want me to do it in this PR or create another issue for that and wait until #118 is merged? I believe this might cause conflicts with #118 since it is modifying the handbook files as well. What do you think?

damithc:
1. Remove the unnecessary div elements you added
2. Combine fragments of one section, to serve as an example

No need to wait for the other PR to merge. Whoever merge second will have to de-conflict :-p Normally, we try to avoid conflicts, but having conflicts gives you a chance to learn how to de-conflict.

jkt001:
I'm doing it now. I'm going to combine the fragments of the policy page into one file as an example.

jkt001:
It is done. What has been done:
<ol>
<li>The policy section of the handbook page has been merged into one file.</li>
<li>All embedded links in the schedule page that refer to the policy section of the handbook page are now able to extract only specific sections of the policy file</li>
</ol>

Editing the policy section of the handbook page is now easier since change is only needed in the policy page and if a new section is added, only need to add the hyperlink to that section on the top of handbook.html, without having to add it into the sections array at the bottom.

Staged version can be viewed <a href = "http://jkt001.github.io/website/">here</a>. It is already up-to-date with the latest master branch.

damithc:
Yes, this is the behavior I wanted. @wkurniawan07 for your review please.

scripts/common.js
```diff
@@ -6,25 +6,35 @@ function makeAccordion(elementSelector) {
...
 }
 
-function getContentUsingAjax(section, elementSelector) {
-    pullContent(section, elementSelector, 'Exract from handbook');
+function getContentUsingAjax(fileName, elementSelector, sectionName) {
+    if (arguments.length == 3) {
+        pullContent(fileName, elementSelector, 'Exract from handbook', sectionName);
+    } else {
+        pullContent(fileName, elementSelector, 'Exract from handbook');
+    }
```
wkurniawan07: I don't think it needs to be split this way. Just pass the function to `pullContent(...)` and later on check that the `sectionName` is undefined rather than the argument length is 4.

scripts/common.js
```diff
@@ -6,25 +6,35 @@ function makeAccordion(elementSelector) {
...
+    if (arguments.length == 4) {
+        var url = fileName +'.html';
+        var target = ' #' + sectionName;
+        var pageToLoad = url + target;
+        var targetSection = elementSelector + sectionName;
+        var newSectionName = sectionName;
+    } else {
+        var pageToLoad = fileName +'.html';
+        var targetSection = elementSelector + fileName;
+        var newSectionName = fileName;
```
wkurniawan07: `embeddedSectionName` is a better name here, `new` is vague.

wkurniawan07: Forget the previous comment, there's more things needing improvement here than that.

scripts/common.js
```diff
@@ -6,25 +6,35 @@ function makeAccordion(elementSelector) {
...
+    }
```
wkurniawan07: This whole code block can be improved. In fact it is possible to remove the whole if/else block to set the variables.

1. Get `var pageToLoad = fileName + '.html';` by default, and when the `sectionName` is defined, append it. It's more logical that way.
2. Define `targetSection` with ternary conditional, based on the presence of the `sectionName` variable. Also you might want to append a delimiter between the `elementSelector` and `*Name`, a `-` should do.
3. `newSectionName` is not useful. The way it is used later on, you can easily replace it with `targetSection.substring(1)`.

scripts/common.js
```diff
@@ -6,25 +6,35 @@ function makeAccordion(elementSelector) {
...
-            $(elementSelector).html('<div><span class="embeddedHeading">' + title + '</span><button onclick="$(\'' + elementSelector + '\').html(\'\');' +
-                                                         ' $(\'' + elementSelector + '\').removeClass(\'embedded\');" ' +
-                                                 'class="btn-dismiss">X</button><br><br></div>' + data);
-            $(elementSelector + ' > div > .btn-dismiss').button();
-        }
-    });
+    $(elementSelector).addClass('embedded');
+    $(elementSelector).html('<div><span class="embeddedHeading">' + title + '</span><button onclick="$(\'' + elementSelector + '\').html(\'\');' +
+       ' $(\'' + elementSelector + '\').removeClass(\'embedded\');" ' +
+       'class="btn-dismiss">X</button><br><br></div> <div id = "' + elementSelector.substring(1) + newSectionName + '"></div>');
```
wkurniawan07: `</div><div id="` (whitespace)

contents/handbook-policy.html
```diff
@@ -0,0 +1,196 @@
+<h3>Policy on following instructions</h3>
+<div id="handbook-policy-followingInstructions">
```
wkurniawan07: You can remove all the leading `handbook-policy-` from the id. It will make much more sense when defining the `targetSection` later (`embedded-week2-policies-followingInstructions` vs `embedded-week2-policies-handbook-policy-followingInstructions`), not to mention you save a large number of characters.

wkurniawan07:
Added some comments. We're close to this being merge-ready.

jkt001:
Alright, working on it right now. I will notify as soon as it is ready for review again.

contents/handbook-policy.html
jkt001: @wkurniawan07 

ok I shall do it. However, in the handbook.html, we need this id to match with the links given on top of the page. Modifying this requires updating the links on top of the handbook.html file, i.e. changing the IDs of the target section since it is changed here. To preserve uniformity, it is better to change for all sections of the handbook instead of just the policy section (it is just removing the 'handbook-' prefix). Should I work on them as well or do I just need to change the policy section in this PR?

wkurniawan07: Don't touch the others, let it be the task of whoever's going to combine the different excerpts (e.g faq, deliverables, ...) into one.

jkt001:
Done. @wkurniawan07 for your review, please.

Staged version is <a href = "http://jkt001.github.io/website/">here</a>.

scripts/common.js
```diff
@@ -20,25 +20,29 @@ function makeAccordion(elementSelector) {
...
+        targetSection += sectionName;
+    }
+
+    $(elementSelector).addClass('embedded');
+    $(elementSelector).html('<div><span class="embeddedHeading">' + title + '</span><button onclick="$(\'' + elementSelector + '\').html(\'\');' +
+       ' $(\'' + elementSelector + '\').removeClass(\'embedded\');" ' +
+       'class="btn-dismiss">X</button><br><br></div> <div id="' + targetSection.substring(1) + '"></div>');
+    $(elementSelector + ' > div > .btn-dismiss').button();
+
+    $(targetSection).load(toBeLoaded);
```
wkurniawan07: Ok I just thought of something. We might not need this new div after all. Here's how: see [how the load API works](http://api.jquery.com/load/), you can work it such that all the processes involving `$(elementSelector)` is done only after load success. I believe you can apply it here.

jkt001:
@wkurniawan07 

Done. I have updated the way of loading such that no extra div is needed.

Staged version is <a href = "http://jkt001.github.io/website/">here</a>. It is combined with PR #138.

wkurniawan07:
Good to go. Thanks for your perseverance :+1:

PR #138, opened 2015-12-31
Schedule page: Expand contents upon clicking subheadings #30

### LINKED ISSUE #30: Schedule page: Expand contents upon clicking subheadings
Revealing the contents upon clicking the subheadings (Deadlines, Things to Do, etc) may work better.
In the common cases of "what is there to do", or "what will happen during the tutorial", it's likely that the student will expand everything under the subheading in the end anyway.
While the current format do give a good summary, more information may be more beneficial than less for the students here.

jkt001:
Fixes #30 

Staged version is <a href = "http://jkt001.github.io/website/">here</a>.

The staged version is combined with PR #133 which is the PR for loading specific section in an embedded link.

### DIFF
--- a/scripts/common.js
+++ b/scripts/common.js
@@ -188,13 +188,30 @@ function loadContent(week) {
                 var component = components[i];
                 makeAccordion('.' + component + '-week' + week);
                 addCollapseAndExpandButtonsForComponents('#' + component + '-content-week' + week, component + '-week' + week);
+               
+                var componentToBeOpened = '#' + component + '-content-week' + week;
+                addAutoExpandSubheadingsBehaviour(componentToBeOpened);
             }
+
             $('.preferences').each(function() {
                 var type = $(this).prop('value');
                 if (!$(this).prop('checked')) {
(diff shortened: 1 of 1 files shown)


jkt001:
Update: I have fixed conflicts after master was updated. Ready for review again. Staged version is <a href = "http://jkt001.github.io/website/">here</a>. Staged version combined with #133

wkurniawan07:
Just wondering, if clicking on the accordion header is the same as clicking the `++` button, then I think it's safe to say that `++` has outlived its usage?

damithc:
Let's keep both. `++` is explicit while clicking on the panel header is implicit. But we should reuse code between the two as much as possible.

jkt001:
The auto-expand behaviour is added such that it "clicks" the '++' button, hence it does not add any code besides adding the behaviour. 

If there's anything I should modify or add, just let me know.

wkurniawan07:
Well the functionality does work. I just can't help the feeling that we should remove the "++" button altogether :p
Anyway, that would be another issue. This is good to go. :+1:

Issue #31, opened 2015-09-30
Add a way to give definitions of terms as tool tips


damithc:
e.g. when the reader hovers over a keyword, it pops up a tool tip containing the definition of the word

jkt001:
I'll work on this issue. 

Just to clarify, what <b>keywords</b> are you expecting to be highlighted? Is it module-related or something else? Perhaps I can do what I can and add tooltips to some keywords and we can add / remove the tooltips to other keywords later on.

damithc:
1. There should be a way to indicate that a certain word has a viewabe definition (e.g. by showing a dashed underline)
2. When the user hovers over (or click), the tool tip appears showing the definition (or more info about the word).  The definition is pulled from another file. 

I suspect there are ready-made solutions for this problem. We just have to find a suitable one.

acjh:
Since we're already using `jquery-ui`, we can use its [`tooltip`](https://jqueryui.com/tooltip/) feature.

PR #140, opened 2015-12-31
Misleading visual representation of links #25

### LINKED ISSUE #25: Misleading visual representation of links
3unnamed
Colours used in links aren't very good.
Blue is the recognised colour for links, but external links are just black and underlined in the content.
Internal links (the blue ones) loads content via AJAX, but it's not clear that it's a link either because hovering over it gives you a vertical bar instead of the usual hand.
Some sort of indicator for AJAX-loading-in-progress would help too; clicking on those give no immediate indication anything even happened.

vhazali:
Fixes #25 


damithc:
The location of the 'loading' gif can be improved. Right now it appears on top of existing content. Instead, it should be at the vertical top of the area where new content is going to appear?
I'm not sure if we should change the link color though. Blue is the common color for links and there is no common color for embedded links. We can think of another way to distinguish between normal links and embedded links. May be use dashed underline? Anyway, that can be done in a separate issue too.

jkt001:
I have a few suggestions:

What I tried to do before was to make it look like this:

screen shot 2015-12-31 at 19 44 29

and when a user hovers over it, it turns dark blue (and the cursor becomes a hand):

screen shot 2015-12-31 at 19 44 51

I don't know if this approach is good, but it's a start. As for dashed underline, in my opinion, it is better for words that have tooltips #31 that are dashed-underline. 

One of my concern for my approach is that it might be confusing with the yellow highlighting as mentioned in #23. Since this is a design type issue, there might be more than one opinion on this. Any other opinion / thoughts are much appreciated.

damithc:
I agree about the dash underline. Let's keep it as before (blue text, underlined, just like a normal link). At a later time, we can add a tiny icon to distinguish embedded links.

styles/common.css
```diff
@@ -367,8 +377,14 @@ table.bordered tr > th {
  */
 
 .embedded-link {
-    color: blue;
+    color: green;
     text-decoration: underline;
+    cursor: pointer;
+    cursor: hand;
```
wkurniawan07: is this double definition of `cursor` intended?

scripts/common.js
```diff
@@ -7,6 +7,7 @@ function makeAccordion(elementSelector) {
 }
 
 function getContentUsingAjax(section, elementSelector) {
+		$("#embedded-link-loading-img").show();
```
wkurniawan07: Any reason why this is put here? Why not in `pullContent` right after `success`, it flows more logically as well.

wkurniawan07:
Take a look at the indentation. Use spaces, not tabs.

wkurniawan07:
Regarding the solution with blue highlight, I don't really suggest it because from HCI point of view highlights are supposedly representing something important, not merely something different.

styles/common.css
vhazali: Yup! For some reasons that I cannot explain, it doesn't work on my chrome if I just put in ```cursor: hand;``` even though some slight googling says that it should

scripts/common.js
vhazali: I wasn't sure if we only wanted to show loading when ```getContentUsingAjax``` was called. I'll edit and move it down to ```pullContent```

vhazali:
Hi all, sorry for the slow reply.

I'll work on the positioning of the image.
As for the colour of the ajax links, should I leave it as green? Or will it be better if I revert it to the original blue?

damithc:
Revert to the original blue.

styles/common.css
```diff
@@ -73,6 +73,16 @@ a:visited {
...
 
+#embedded-link-loading-img {
+  display: none;
+  position: fixed;
+  top: 50%;
+  left: 50%;
+  z-index: 11;
+  background-color: white;
+  font: 1.5em;
+}
```
wkurniawan07: 4 spaces of indentation, not 2.

styles/common.css
```diff
@@ -369,6 +379,12 @@ table.bordered tr > th {
 .embedded-link {
     color: blue;
     text-decoration: underline;
+    cursor: pointer;
+    cursor: hand;
+}
+
+.embedded-link:active {
+    color:purple;
```
wkurniawan07: `color: purple;`

contents/schedule.html
```diff
@@ -7,6 +7,7 @@
         <link type="text/css" rel="stylesheet" href="../styles/common.css" />
     </head>
     <body>
+        <img id="embedded-link-loading-img" src="/images/ajax-preload.gif" alt="Loading..."/>
```
wkurniawan07: Two issues:
1. Don't close the `img` tag with `/>`.
2. The reference is broken. It should be `../images/ajax-preload.gif`. `common.js` is making the same mistake (but don't fix it here).

styles/common.css
wkurniawan07: What if you just put in `cursor: pointer;`? If it still fails, then I guess we have no choice but to include both.

styles/common.css
```diff
@@ -369,6 +379,12 @@ table.bordered tr > th {
 .embedded-link {
     color: blue;
     text-decoration: underline;
+    cursor: pointer;
+    cursor: hand;
```
wkurniawan07: You mentioned that putting in only `cursor: hand;` fails. What if you put in only `cursor: pointer;`? If it still fails, then I guess we have no choice but to include both.

wkurniawan07:
Some minor changes and we're good to go if @damithc agrees with the solution :+1:

damithc:
When I check the staging site, the 'loading gif' never shows up. All I see is the default 'image missing' placeholder. Is the image missing? Or the loading of the 'loading gif' takes too long (how ironic :-p)

damithc:
Is the gif problem due to #146 ?

vhazali:
@damithc yes I think it's due to the '''src''' tag as pointed out by @wkurniawan07 

I'll get to fixing the problems in awhile! 

I still can't find a good way for the loading gif to appear right under the link that's clicked. Might need a bit more time for that

damithc:
Fine tuning the position can be done as a separate issue after this. Just create a new issue for it (and for any other potential improvements) after this one is merged.

styles/common.css
vhazali: Using just ```cursor: pointer;``` works!

wkurniawan07:
Good to go :+1:

PR #144, opened 2016-01-02
Add a "Back to Top" button on the handbook page. #141

### LINKED ISSUE #141: Add a "Back to Top" button on the handbook page.
Currently, in the handbook page, we can navigate to a specific section (such as the appendix). However, once a user clicks the link to a section, he/she cannot navigate easily back to the top of the page. The user needs to scroll all the way to the top. 

It would be nice if a  "back to top" button that scrolls with the user (i.e. always clickable wherever the user is currently at) that can take the user back to the list of contents of the handbook.

HooDeLin:
Fixes #141 
The staging site is [here](http://hoodelin.github.io/website/)


contents/handbook.html
```diff
@@ -12,83 +12,85 @@
...
 <div id="modal">Page Loading...</div>
-<ul>
-    <li><a href="#handbook-preliminaries">Preliminaries</a></li>
-    <li><a href="#handbook-lectures">Lectures</a></li>
-    <li><a href="#handbook-tutorials">Tutorials</a></li>
-    <li><a href="#handbook-textBooks">Text Books</a></li>
-    <li><a href="#handbook-programmingLanguages">Programming Language</a></li>
-    <li><a href="#handbook-project">Project</a></li>
+<div id="back-to-top-button">Back To Top</div>
+<div id="table-of-contents">
```
damithc: Is this change part of this PR?

HooDeLin: I have wrapped all of the contents inside a `<div>` and indented them

styles/handbook.css
```diff
@@ -33,4 +33,21 @@ code.clisystem{
...
+#back-to-top-button{
+	display: block;
+    position: fixed;
+    bottom: 0;
+    right: 0;
+    padding: 10px;
+    border: none;
+    color: white;
+    text-shadow: -1px 1px #417cb8;
+    background: #0fcbb4;
```
wkurniawan07: The indentation rule is 4 spaces, but since it's not followed in this entire file, use tabs for everything. We'll get someone else to fix it afterwards.

contents/handbook.html
```diff
@@ -12,83 +12,85 @@
...
+            <li><a href="#handbook-policy-teamSize">Policy on grading smaller/larger teams</a></li>
+            <li><a href="#handbook-policy-workDistribution">Policy on project work distribution</a></li>
+            <li><a href="#handbook-policy-validAbsences">Policy on absence due to valid reasons (e.g. MC, LOA, University events)</a></li>
+            <li><a href="#handbook-policy-responseTime">Policy on email response time</a></li>
+            <li><a href="#handbook-policy-techHelp">Policy on tech help</a></li>
+            <li><a href="#handbook-policy-publishingSubmissions">Policy on publishing submissions</a></li>
+            <li><a href="#handbook-policy-plagiarism">Policy on plagiarism</a></li>
+            <li><a href="#handbook-policy-reuse">Policy on reuse</a></li>
+            <li><a href="#handbook-policy-outsiderHelp">Policy on help from outsiders</a></li>
+            <li><a href="#handbook-policy-submissionLength">Policy on suggested length for submissions</a></li>
```
wkurniawan07: Just a heads up that you'll see conflicts here once #133 is merged.

scripts/handbook.js
```diff
@@ -23,3 +23,33 @@ $('a').click(function() {
...
+function isTableOfContentVisible() {
+    var windowTop = $(window).scrollTop();
+    var tableTop = $('#table-of-contents').offset().top;
+    var tableBottom = $('#table-of-contents').height() + tableTop;
+    return windowTop < tableBottom;
+}
+
+function makeBackToTopButton(buttonAnimationDuration) {    
+    var speed = 1;
+    $("#back-to-top-button").css("display", "none");
```
wkurniawan07: You can move this attribute to the CSS file, replacing the `block` (i.e the display is `none` by default). `fadeIn` and `fadeOut` will change it on demand.

wkurniawan07: Also, add this line `$("#back-to-top-button").button()`.

scripts/handbook.js
```diff
@@ -23,3 +23,33 @@ $('a').click(function() {
...
+    $('#back-to-top-button').click(function(e) {
+        $("#back-to-top-button").fadeOut(buttonAnimationDuration);
```
wkurniawan07: I don't think you need this. When the page is scrolled afterwards, the `$(window).scroll` events should be automatically triggered.

scripts/handbook.js
```diff
@@ -23,3 +23,33 @@ $('a').click(function() {
...
+        $('html, body').animate({
+            scrollTop: 0
+        }, speed);
+    });
+}
+
+$(document).ready(function() {
+    var buttonAnimationDuration = 500;
```
wkurniawan07: 500 is too slow for a button to appear. Try a lower value.

styles/handbook.css
wkurniawan07: Also, the choice of color is unusual. Can you pick one of the colors used in the schedule page's innermost accordion? The green one used for tutorials might be better.

scripts/handbook.js
```diff
@@ -23,3 +23,33 @@ $('a').click(function() {
...
+    makeBackToTopButton(buttonAnimationDuration);
+    $(window).scroll(function() {
+        if (!isTableOfContentVisible()) {
+            $('#back-to-top-button').fadeIn(buttonAnimationDuration);
+        } else {
+            $('#back-to-top-button').fadeOut(buttonAnimationDuration);
+        }
+    })
+});
```
wkurniawan07: I think the pre-existing function `$('a').click` can be part of this `$(document).ready`. You can spell out the entire `makeBackToTopButton` without the need for wrapping it as a function, then you can re-use the variable `speed` for the `$('a').click` as well.

wkurniawan07:
Added some comments for improvement. The functionality is there and working good :+1:

HooDeLin:
The improvements are made.

styles/handbook.css
```diff
@@ -46,4 +46,21 @@ img {
...
+	padding: 10px;
+	border: #dff0d8;
+	color: #3c763d;
+	background: #dff0d8;
+	transition: background 0.2s ease;
+}
+
+#back-to-top-button:hover{
+	color: #fff;
+	background: #5cb85c;
```
wkurniawan07: Small request: interchange the color scheme? Right now the non-hovered version is light-colored on white background, which is not really noticeable. If the darker green color is used as the default the button's presence will be more obvious.

scripts/handbook.js
```diff
@@ -14,12 +14,37 @@ for (var i in sections) {
...
+        }, speed);
+    });
+
+    $(window).scroll(function() {
+        if (!isTableOfContentVisible()) {
+            $('#back-to-top-button').fadeIn(buttonAnimationDuration);
+        } else {
+            $('#back-to-top-button').fadeOut(buttonAnimationDuration);
+        }
+    })
```
wkurniawan07: Close with `;`

scripts/handbook.js
```diff
@@ -14,12 +14,37 @@ for (var i in sections) {
...
+            scrollTop: 0
+        }, speed);
+    });
+
+    $('a').click(function() {
+        var adjustment = 73;
+        var header = this.hash;
+        $('html,body').animate({
+            scrollTop: $(header).offset().top - adjustment
+        }, speed);
```
wkurniawan07: Let's take it one step further: you can abstract out this `animate` function (used twice as you can see), by supplying just the `scrollTop` parameter. The `speed` variable can then be a local variable in that abstracted out function.

styles/handbook.css
```diff
@@ -1,6 +1,6 @@
 body {
-    padding: 0px 10px;
-    max-width: 740px;
+	padding: 0px 10px;
+	max-width: 740px;
```
wkurniawan07: You don't need to fix this. Unless they're added by you or related to your work, let the non-conforming styles be.

wkurniawan07:
Minor changes requested and we should be good to go afterwards.

HooDeLin:
The requested changes are done

wkurniawan07:
Good to go. Once again, well done :+1:

Issue #116, opened 2015-12-25
Handbook: instructions about the handout is incorrect


damithc:
It says handouts are uploaded every Wed. It should say that the full handout is uploaded to IVLE workbin at the beginning of the semester.

PR #147, opened 2016-01-03
Handbook: instructions about the handout is incorrect #116

harishv7:
Fixes: #116 
Here is a preview: http://harishv7.github.io/website-1/

New description: Each lecture consists of 1-3 sections, supported by
separate handouts. The full handbook containing all the sections is
uploaded to IVLE workbin at the beginning of the semester.
You are welcome to, but not required to, read through them before
attending the lecture.


damithc:
See if you can squash the commits. One commit should be enough for a minor change like this one.

harishv7:
Squashed.

PR #151, opened 2016-01-05
Search page not loaded correctly #150

### LINKED ISSUE #150: Search page not loaded correctly
Search page can be loaded correctly on local server.
On remote server, it might not be loaded correctly because data collection process is slower and might not have completed when data processing began.

zeulb:
Fixes #150 

Add timeout before data processing began to let data collection completed first.

Preview: http://zeulb.github.io/website/


wkurniawan07:
Should be good to go.

PR #152, opened 2016-01-05
Problems with scroll section to view #135

### LINKED ISSUE #135: Problems with scroll section to view
I believe this is due to #123 

The scrolling works in most scenarios. However, if we expand using `+ + + +` and close it, and expand one of the weeks using `+ + +`, it automatically scrolls to the end and prevent the users to scroll up for quite some time before returning normal.

zeulb:
Fixes #135 

This problem happens because multiple scroll animation triggered when clicking + + + + button. 

Preview: http://zeulb.github.io/website/

### DIFF
--- a/scripts/common.js
+++ b/scripts/common.js
@@ -240,11 +240,17 @@ function getDate(week, day) {
 }
 
 function addAutoScrollToClickedWeekHeader() {
+    var isAnimating = false;
     $('.buttoned').click(function(event) {
+        // Don't scroll when an animation is in progress
+        if (isAnimating) return;
         var scrollTarget = '#' + event.currentTarget.id;
+        isAnimating = true;
         $('html, body').animate({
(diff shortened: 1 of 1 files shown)


scripts/common.js
```diff
@@ -240,11 +240,17 @@ function getDate(week, day) {
 }
 
 function addAutoScrollToClickedWeekHeader() {
+    var isAnimating = false;
     $('.buttoned').click(function(event) {
+        // Don't scroll when an animation is in progress
+        if (isAnimating) return;
```
wkurniawan07:
```javascript
if (isAnimating) {
    return;
}
```

wkurniawan07:
Other than that one minor change this is fine :+1: I'll just set as `toMerge` first.

zeulb:
Alright, I have made the change.

PR #154, opened 2016-01-06
Wrong reference of ajax-preload.gif in common.js #146

### LINKED ISSUE #146: Wrong reference of ajax-preload.gif in common.js
In common.js, the reference to <code>ajax-preload.gif</code> is broken. 

The image is located in the images folder whereas <code>common.js</code> is in the scripts folder. The way of referencing it right now will always result in a file not found error.

vhazali:
Fixes #146


damithc:
BTW, any idea why the image shows up with a white background (when used from the Schedule page) although it is supposed to have a transparent background?

damithc:
As for positioning of the gif, may be replace the content of the target div with the gif? i.e. the gif appears exactly where the new content will appear. Currently, the gif appears on top of the existing content of the page.

vhazali:
I set the background to be white. In case the image/ text loads on top of certain colours that make the image/words unclear. 

I'll work on the positioning of the gif for now

PR #159, opened 2016-01-10
Improve text layout on mobile device #155

### LINKED ISSUE #155: Improve text layout on mobile device
For any smartphone whose view is 500px or below.

- Currently only about 2-4 words for bullet points

- Table text shown in excessively large font-size

acjh:
Fixes #155
<a href="http://acjh.github.io/website">Preview</a>

| Before | After |
| -------- |--------|
|  |  |
|  |  |
|  |  |
|  |  |

### DIFF
--- a/styles/common.css
+++ b/styles/common.css
@@ -446,6 +446,15 @@ img {
         height: 100% !important;
         padding-top: 0px !important;
     }
+
+    .ui-accordion-content {
+        padding-left: 10px !important;
+        padding-right: 10px !important;
+    }
+
+    ul {
+        padding-left: 20px !important;
+    }
 }
 
 /*
--- a/styles/common.css
+++ b/styles/common.css
@@ -434,6 +434,15 @@ img {
     }
 }
 
+/* For any device whose view is 500px or below */
+@media screen and (max-device-width: 500px) {
+    body {
+        margin-right: 2%;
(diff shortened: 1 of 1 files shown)


styles/common.css
```diff
@@ -434,6 +434,33 @@ img {
...
+        padding-right: 10px !important;
+    }
+
+    ul {
+        padding-left: 20px !important;
+    }
+
+    table.bordered tr > td {
+        font-size: 12px !important;
+    }
```
wkurniawan07: Actually, the oversized table text size is an oversight on my part when I did #130. The font size should be 14px which is the same as the others. I'll have it fixed right up so you can remove this definition, or is there another reason to make it even smaller?

wkurniawan07:
Good work :+1: just that one minor issue.

styles/common.css
acjh: 12px for table text seems better than 14px on mobile, as it is still large enough but easier to scan for information, as well as reduces scrolling. Limited observation though, as the example above appears to be the only instance of that class.

wkurniawan07: Ok, we can let this in for now, later on when I fix up my mistake in #130 I'll see which alternative is better. Good job on this :+1:

acjh: Thanks!

Issue #157, opened 2016-01-10
C++ coding standard, Dennis Ritchie name should be under the photo


damithc:
Currently, it can appear on the right of the photo.

![image](https://cloud.githubusercontent.com/assets/1673303/12219478/71a8b57a-b780-11e5-8f8a-49c0a3a051c1.png)

acjh:
Note that this issue no longer appears after a max-width was set.
The intended fix is to add <code>\<br\></code> as done in <a href="http://acjh.github.io/website/contents/coding-standards-java.html">coding-standards-java</a>.

acjh:
We no longer have a C++ coding standard.

PR #161, opened 2016-01-10
Style guides: set a max width for the page content #158

### LINKED ISSUE #158: Style guides: set a max width for the page content
It is better to set a max width for the page. Otherwise the content become hard to read when using a wide screen browser.

acjh:
Fixes #158
Previews: <a href="http://acjh.github.io/website/contents/coding-standards-cpp.html">cpp</a>, <a href="http://acjh.github.io/website/contents/coding-standards-html.html">html</a>, <a href="http://acjh.github.io/website/contents/coding-standards-java.html">java</a> coding standards


wkurniawan07:
Good job :+1:
I won't say that this is the actual fix for #157 though, because it _just happens_ to fix it. The actual fix will be adding a `<br>`, just like what is done for the Java coding style. So this PR shouldn't close that issue.

damithc:
I thought the page width is a css issue? Using <br> to control pad width is going to be quite messy and fragile.

wkurniawan07:
Ah no that's not relevant to this issue. It's about #157 the Dennis Ritchie text on the side of the picture, clearly caused by a missing `<br>`.

damithc:
Understood :-)

acjh:
Noted, but I suggest to specify on #157 since that issue won't seem obvious anymore after this PR.

damithc:
Good idea @acjh. Can you add a comment there?

acjh:
Done!

Issue #145, opened 2016-01-02
Typo in common.js


vhazali:
`function getContentUsingAjax(section, elementSelector) {
    pullContent(section, elementSelector, 'Exract from handbook');
}`

Should be ``Extract from handbook`);`

Affects all embedded-links
![Exract typo image](https://cloud.githubusercontent.com/assets/8543547/12074480/25b9640e-b189-11e5-96b8-70cb62609d5c.png)

PR #172, opened 2016-03-17
Issue #145: Fixed typo 'exract' in common.js, changed to 'extract'

cxlim:
In function getContentUsingAjax in the file Common.js, there was a typo 'Exract from handbook', which affects all the embedded handbook links.

Fixed by changing to `Extract from handbook`.

### DIFF
--- a/scripts/common.js
+++ b/scripts/common.js
@@ -21,7 +21,7 @@ function makeAccordion(elementSelector) {
 }
 
 function getContentUsingAjax(fileName, elementSelector, sectionName) {
-    pullContent(fileName, elementSelector, 'Exract from handbook', sectionName);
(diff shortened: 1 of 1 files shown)


damithc:
@wkurniawan07 for your review and merging.

acjh:
Fixed in #196. @damithc Any idea why GitHub says this branch has no conflicts with the [base branch](https://github.com/nus-cs2103/website/blob/master/scripts/common.js#L24)? :O

damithc:
Possibly because the two patches are _exactly_ the same?

damithc:
@cxlim As it turns out, the bug is already fixed by another PR. Nevertheless, thanks for the contribution.

PR #173, opened 2016-03-19
Changed the highlighting in the headers to a less jarring yellow

cxlim:
Classified highlighted headers under a new class called 'highlighted-headertext', with a paler yellow as background.

### DIFF
--- a/contents/week1.html
+++ b/contents/week1.html
@@ -8,7 +8,7 @@
             </div>
         </div>
         <div class="things-to-do things-to-do-week1">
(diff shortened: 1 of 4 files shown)


acjh:
Fixes #23

| Before | After |
| --- | --- |
|  |  |

@damithc Actually, I agree with your comment in #184 that the current highlight doesn't look that bad. Should we close this issue?

damithc:
Yes @acjh , let's keep the current one. We should use highlighting sparingly in headers, but when we do, I'd rather have it bright and eye-catching. :-)

damithc:
@cxlim thanks for the PR. Sorry we took so long to process it.

PR #174, opened 2016-04-10
Refactor load content in context

anaprimawaty:
Removed inline JS to make code more readable and reusable

addEmbeddedLinkBarModal() in common.js appends an empty div (i.e. <div></div>) to each embedded-link-bar, replacing the need for an empty div (i.e. <div id="embedded-week1-general"></div>) in the html files.

Refactoring is only applied to week1.html for demonstration. Will apply to the rest if approach is ok!

Here is a [preview](http://anaprimawaty.github.io/website/)

### DIFF
--- a/contents/week1.html
+++ b/contents/week1.html
@@ -38,10 +38,10 @@
                 <b>Handbook sections to read around this time:</b>
                 <ul>
                     <li>
-                        General: <span onclick="getContentUsingAjax('handbook-preliminaries', '#embedded-week1-general');"><span class="embedded-link">Preliminaries</span></span> | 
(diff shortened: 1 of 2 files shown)


damithc:
In the original code, there is a way to specify which location to load the new content (it may not always be right below the link). Do you have the same feature in new code?

anaprimawaty:
Now able to load new content in a specified location.

Also able to support embedded links within embedded content.

You can view the two implementations by following the red tags at Week 1 > Things to do > Have a look at the 'module handbook'.

acjh:
@anaprimawaty I like what's being done here! Do you want to tidy up this PR?

PR #184, opened 2016-08-11
Yellow highlight looks jarring on blue subheadings #23

### LINKED ISSUE #23: Yellow highlight looks jarring on blue subheadings

acjh:
Fixes #23
[Preview](https://cdn.rawgit.com/acjh/website/23-yellow-highlight-jarring/)

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/contents/week1.html
+++ b/contents/week1.html
@@ -61,7 +61,7 @@
       </div>
     </div>
     <div class="things-to-do things-to-do-week1">
-      <h3>Attend Lecture 1 <span class="highlighted-darktext">(compulsory)</span></h3>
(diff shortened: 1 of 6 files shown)


damithc:
It's not about header, but about the background color. We should have different highlight styles that we can choose based on the context. 
On a related note, I think the current dark text with white background doesn't look that bad. Highlighting is supposed to be eye catching anyway. 
Keep it as it is and close the issue?

damithc:
Alternatively, we can try a different shade of yellow that doesn't look jarring with dark backgrounds.

PR #200, opened 2016-12-06
Add popups to images for zoom effect

jia1:
screenshot_2016-12-06-21-53-05

Popped-up images are set to fit the width of the screen.


damithc:
Ready for review?

jia1:
Yes. Apologies for leaving out the indication of ready to review.

damithc:
@acjh can review this one?

contents/handbook.html
```diff
@@ -257,6 +258,7 @@
 
 </script>
 <p>-------------------------------------End of document----------------------------------------</p>
+<script type="text/javascript" src="../vendor/magnific-popup/jquery.magnific-popup.min.js"></script>
```
acjh: Currently, popups are initialised in `common.js`, which handbook doesn't load.
But some considerations before you rush to initialise popups in `handbook.js`:
- Cloning code is generally bad.
- Handbook content now loads in a different manner - you may have to iron out those bugs.
- Anyway, there are no images in handbook worth popping up, that's not already a good size.

I'd recommend targeting the schedule page for now, which is sufficient (and what was intended).

styles/common.css
```diff
@@ -540,3 +540,45 @@ img {
 .separator {
     border-top: 1px solid #ccd8e2;
 }
+
+/*
+    CSS for magnific-popup
+*/
+
+/* padding-bottom and top for image */
+.mfp-no-margins img.mfp-img {
```
acjh: `.mfp-no-margins` is an unused class.

acjh:
@jia1 Note that popup `!=` zoom effect, which in the case of `magnific-popup` refers to animation.
Currently, there's no zoom effect (see Week 5 in [my earlier attempt](https://rawgit.com/acjh/website/168-enable-image-popup-mobile/) if unclear), but it's not required.

PR #202, opened 2016-12-08
Fix size and position of 'Page Loading...' overlay in handbook page #112

### LINKED ISSUE #112: Fix size and position of 'Page Loading...' overlay in handbook page

acjh:
Fixes #112
[Preview](http://rawgit.com/acjh/website/112-fix-loading-overlay-handbook/)

1. **Fixed overlay not appearing**
I noticed that `handbook.html` was blocking UI updates while it was loading into the `content` frame.
After implementing this fix using `setTimeout`, I found [this explanation for its use](http://stackoverflow.com/questions/779379/why-is-settimeoutfn-0-sometimes-useful#779785) on Stack Overflow.

2. **Fixed size and position of overlay (on mobile)**

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/contents/handbook.html
+++ b/contents/handbook.html
@@ -12,8 +12,7 @@
 </head>
 <body>
 
-<div id="overlay"></div>
-<div id="modal">Page Loading...</div>
+<div id="overlay">Page Loading...</div>
 <div id="back-to-top-button">Back To Top</div>
 <div id="table-of-contents">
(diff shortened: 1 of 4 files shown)


acjh:
@damithc I moved the overlay removal back after the synchronous block to avoid the unnecessary checks.

PR #216, opened 2016-12-13
Handbook TOC links are not working in iPhone browser #19

### LINKED ISSUE #19: Handbook TOC links are not working in iPhone browser

acjh:
Fixes #19
[Preview](http://rawgit.com/acjh/website/19-handbook-links-in-iphone/)

### DIFF
--- a/scripts/handbook.js
+++ b/scripts/handbook.js
@@ -4,13 +4,15 @@
  * Account for Safari scrolling.
  */
 function jumpToSectionHeading(section) {
-    var headerPosition = $('#' + section).prev().offset().top;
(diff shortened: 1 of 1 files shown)


acjh:
As it turns out, Apple doesn't like creating compatible products.
I managed to get my hands on an iPad just now and noticed that Safari doesn't scroll *within* frames.
The solution is to make `window.top` scroll instead.

Note: Actually, this problem is not limited to `handbook` TOC links, but **all** anchor links within frames.

damithc:
Works for me in iPhone6/Safari

acjh:
Shall I merge this?

damithc:
Yes, go ahead.

PR #220, opened 2016-12-14
Add a way to give definitions of terms as tooltips #31

### LINKED ISSUE #31: Add a way to give definitions of terms as tool tips

acjh:
Fixes #31
[Preview: Week 0](http://rawgit.com/acjh/website/31-add-way-definition-tooltips/contents/week.html?preview=0)

This is still a proof-of-concept, suggestions welcome :)

> A hyperlink in a `tooltip` is **not** meant to be clicked, so don't get frustrated over trying to click it.
> It's just there to demo HTML content. But yes, it works.

### DIFF
--- a/scripts/tooltip.js
+++ b/scripts/tooltip.js
@@ -1,5 +1,5 @@
-// Maps a string to a definition (in the form of a HTML string)
-var definitions = {
+// Maps a string to a tooltip (in the form of a HTML string)
+var tooltips = {
 };
 
 // Maps a string to a list of aliases
@@ -7,18 +7,18 @@ var aliases = {
 }
 
 for (let term in aliases) {
-    let definition = definitions[term];
+    let tooltip = tooltips[term];
     let values = aliases[term];
     for (let i in values) {
-        alias = values[i];
-        definitions[alias] = definition;
+        let alias = values[i];
+        tooltips[alias] = tooltip;
     }
 }
 
 var count = 0;
 $('tooltip').each(function() {
     var term = $(this).text();
-    var title = definitions[term];
+    var title = tooltips[term];
     var id = String(count++);
     $(this).attr('id', id);
     $(this).tooltip({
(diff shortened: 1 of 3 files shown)


damithc:
This is good, but perhaps we should also cater for cases where the term can be given in different ways?
One option: `<tooltip>Refactoring</tooltip>`, `<tooltip term='Refactoring'>Refactor</tooltip>`
Another option is to supply all aliases where we define the terms.

acjh:
+1 for supplying all aliases where we define the terms, which would mean less clutter in the HTML :)
It's also easier to maintain (and to swop the 'key') by keeping track of all tooltip data in a single file.

scripts/tooltip.js
```diff
@@ -0,0 +1,32 @@
+// Maps a string to a definition (in the form of a HTML string)
+var definitions = {
```
damithc: `tooltips`? We don't have to limit this to definitions only.

damithc:
The downside is you cannot have multiple definitions for the same keyword. Sometimes the same term can have different definitions depending on the context. Also there can be a simple definition and a longer definition. So the ideal is to be able to do both ways. But this is good enough for now. Adding the other method can be a separate issue.

acjh:
Ready for review.

damithc:
Can merge. Create an issue for the enhancement we discussed above?

damithc:
@acjh 
Can you give an example of how to add a tooltip (including aiases)? I think the current code doesn't have any examples to follow. You can give the example here in this issue thread and I'll follow that to add a few tooltips.

acjh:
Oh, I didn't want to pull the demo.
See [this commit](https://github.com/acjh/website/commit/d5c6524d3e2363418d25cc7149a0c6d73c6d1ad4).

damithc:
Got it. 👍

PR #244, opened 2017-01-12
Schedule: keep week 0 expanded during week 1 #243

### LINKED ISSUE #243: Schedule: keep week 0 expanded during week 1
Students can miss week 0 content if only week 1 is expanded in the first week. May be we treat week 0 as a special case and expand it during week 1 too?

acjh:
Fixes #243 
[Preview](http://rawgit.com/acjh/website/243-keep-week-0-expanded/)

### DIFF
--- a/scripts/common.js
+++ b/scripts/common.js
@@ -242,7 +242,9 @@ function loadContent(week) {
                 expandWeekFully(week);
 
                 if (week == 1) {
-                    expandWeekFully(0);
+                    setTimeout(function() {
+                        expandWeekFully(0);
+                    }, 500); // Wait for scroll animation to complete for Week 1
                 }
             }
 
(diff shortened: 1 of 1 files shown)


scripts/common.js
```diff
@@ -240,6 +240,12 @@ function loadContent(week) {
                 totalWeeks = 1;
             } else if (isCurrentWeek(week)) {
                 expandWeekFully(week);
+
+                if (week == 1) {
```
damithc: May be add a comment to explain this spacial case?

acjh:
@damithc Shall I merge?

damithc:
> Shall I merge?

Yes, go ahead 👍

damithc:
@hwkchia FYI

Issue #225, opened 2016-12-18
Style embedded links differently from normal links


damithc:
Embedded links (those that expand content in the page itself) are different from normal links. It's good if we can style them in a slightly different way from normal links.

PR #252, opened 2017-03-06
Style embedded links differently from normal links #225

MasterFelix15:
Fixes issue: #225 

Staging site: https://rawgit.com/MasterFelix15/website/225-differentiate-embedded-links/index.html

The website now displays the embedded links in dark blue and in italic.

### DIFF
--- a/styles/common.css
+++ b/styles/common.css
@@ -432,7 +432,7 @@ table.two-column-content td {
  */
 
 .embedded-link {
-    color: blue;
+    color: darkblue;
     text-decoration: underline;
     font-style: italic;
     cursor: pointer;
--- a/styles/common.css
+++ b/styles/common.css
@@ -434,6 +434,7 @@ table.two-column-content td {
 .embedded-link {
     color: blue;
     text-decoration: underline;
+    font-style: italic;
     cursor: pointer;
 }
 


acjh:
Deferring to @damithc. I couldn't find a good reference on styling embedded/ajax links, so I'm open to this.

damithc:
I'm not very keen on using _italics_ because we use it for other things (e.g. when introducing a new term).
I remember some sites using a small arrow icon to indicate a link 'opens in new window'. Something similar is possible to indicate a link 'expands'?. Also, I think we use light gray dotted underlines to indicate tool tips. Can we use a variation of that to indicate this type of links?

MasterFelix15:
@damithc @acjh I have changed the implementation to include a small icon which indicates that the link is expandable.

damithc:
> I have changed the implementation to include a small icon which indicates that the link is expandable.

Good work. How about special underline? Is that possible? If it is, we can also remove the blue color.

MasterFelix15:
@damithc @acjh Added dotted underline and removed blue colour.
screen shot 2017-03-05 at 11 49 52 pm

damithc:
> I have changed the implementation to include a small icon which indicates that the link is expandable.

Looks good.

styles/common.css
```diff
@@ -432,11 +432,16 @@ table.two-column-content td {
  */
 
 .embedded-link {
-    color: blue;
-    text-decoration: underline;
+    color: black;
```
damithc: Perhaps we should let it keep the original color rather than override to black?

MasterFelix15:
@damithc @acjh Colour changed to original blue.
screen shot 2017-03-06 at 9 21 25 am

damithc:
> Colour changed to original blue

I meant retain the color of the text (i.e. don't override the color at all).

MasterFelix15:
@damithc @acjh Sorry I misunderstood. The colour override has been cancelled.
screen shot 2017-03-06 at 5 13 22 pm

damithc:
Looks good @MasterFelix15. A slight tweak would be to ensure the underline and the icon are the same color (grey, or light blue) but not strictly necessary.

MasterFelix15:
@damithc @acjh The colour of the icon has been adjusted to grey to be consistent with the underline. 
screen shot 2017-03-06 at 7 22 47 pm

acjh:
Hmm, the icon looks really pixelated. How about using CSS (sprite sheet) from FontAwesome (30 KB)? In comparison, the current icon is just 300 B.

Alternatives:
- jQuery UI (already loaded for ui-accordion's triangle)
- Bootstrap (more than 100 KB)

jQuery UI's icon seems a bit small (can try), while Bootstrap comes with a lot of unused styles (that I previously had problems using with jQuery UI).

damithc:
@MasterFelix15 any updates?

MasterFelix15:
@damithc My apologies professor, I am having my exam week here in UCSD, can you allow me a few more days?

damithc:
> @damithc My apologies professor, I am having my exam week here in UCSD, can you allow me a few more days?

Sure.

MasterFelix15:
@damithc @acjh Sorry for not being active for so long, I am having my exam week in UCSD, and it still isn't over yet... 

Regarding the previous problem, I looked into Fontawesome and there seem to be a small license issue if I want to use their icons. Besides, I do not think it is necessary to bring in an entire library just for that icon we want to use.

I have explored a bit and found out that there is really a hacky way of doing this: using the after selector in CSS and set the icon image as its background image. If we set the image as the content, we wouldn't be able to adjust the display size of the icon image, however, if we use it as a background image, we can do that by adjusting the background size.
screen shot 2017-03-19 at 11 50 43 pm

styles/common.css
```diff
@@ -432,11 +432,23 @@ table.two-column-content td {
  */
 
 .embedded-link {
-    color: blue;
-    text-decoration: underline;
+    border-bottom: 1px dashed #999;
+    text-decoration: none;
```
acjh: Can remove `text-decoration` entirely (i.e. don't override too)

styles/common.css
```diff
@@ -432,11 +432,23 @@ table.two-column-content td {
...
+    width: 10px;
+    height: 10px;
+    display: inline-block;
+    background: url("../images/expand.png");
+    -webkit-background-size: 10px 10px;
+    background-size: 10px 10px;
+    background-repeat: no-repeat;
+    background-position: 75% 75%;
+}
+
```
acjh: - Let's place this after `.embedded-link:active`
- Use `::after`
- Order properties alphabetically, `-webkit` first

acjh: - `content` can be an empty `""`
- You can use `padding-left: 1ex;` for a little space before the icon

acjh: @MasterFelix15 Any updates?

Also, the `-webkit` property is unnecessary.
Chrome 4.0 and Safari 4.1 (before which the `-webkit` prefix was used) were released in 2010.

MasterFelix15:
@damithc @acjh Added small space before the icon, changed the space from 1ex to 0.5ex to look more comfortable. Cleaned up code as instructed, however, the background no-repeat element has to remain under ::after in order for it to work properly. And the padding-left has to be changed to margin-left since we are trying to control the background picture.
screen shot 2017-04-05 at 10 47 34 am

styles/common.css
```diff
@@ -432,13 +432,24 @@ table.two-column-content td {
...
-    color: blue;
-    text-decoration: underline;
+    border-bottom: 1px dashed #999;
     cursor: pointer;
 }
 
+.embedded-link::after {
+    background: url("../images/expand.png");
+    background-size: 10px 10px;
+    background-repeat: no-repeat;
```
acjh: `background-repeat` before `background-size`

styles/common.css
```diff
@@ -432,13 +432,24 @@ table.two-column-content td {
...
+    content: "";
+    display: inline-block;
+    height: 10px;
+    margin-left: 0.5ex;
+    width: 10px;
+}
+
 .embedded-link:active {
     color: purple;
+    background-position: 75% 75%;
```
acjh: I think you misinterpreted this.
I meant moving the whole `.embedded-link::after` below `.embedded-link:active`.

That said, this shows that `background-position` is not necessary.
Could you also confirm that `background-repeat` is not necessary?

acjh:
@MasterFelix15 Any updates?

